PR #644, opened 2015-09-16
Issue#650

daveymoran-nearform:


web/public/js/controllers/dojo-event-form-controller.js
```diff
@@ -347,10 +347,19 @@
...
         $scope.eventInfo.userType = _.where($scope.eventInfo.userTypes, {name: $scope.eventInfo.userType})[0];
+        $scope.pastEvent = isEventInPast(_.last(event.dates));
+
         done(null, event);
       }, done);
     }
 
+    function isEventInPast(dateObj) {
+      var now = moment.utc().valueOf();
+      var end = moment.utc(dateObj.endTime).valueOf() - (moment().utcOffset() * 60000);
```
ckiss: you can do here something like moment.utc().subtract(utc_offset_value, 'minutes') search in the code for places where I've done this already. manage-events-controller should have already a similar function.

PR #649, opened 2015-09-16
update message.po for dojo stage values

kennanseno:
https://github.com/CoderDojo/community-platform/issues/636


daveymoran-nearform:
lgtm :)

PR #651, opened 2015-09-17
Fix event start time

david-cahill:


web/public/js/controllers/dojo-event-form-controller.js
```diff
@@ -73,6 +70,12 @@
     $scope.datepicker = {};
     $scope.datepicker.minDate = now;
 
+    $scope.$watch('eventInfo.date', function (startDate) {
+      var newStartDate = angular.copy(startDate);
```
ckiss: does this actually change the startdate of the event? cause I can see it changes the endTime, not the startTime though

web/public/js/controllers/dojo-event-form-controller.js
```diff
@@ -159,7 +162,7 @@
         // Extend eventInfo
         if(eventInfo.recurringType === 'weekly') {
           eventInfo.dates = getEveryTargetWeekdayInDateRange(
-            eventInfo.startTime,
+            eventInfo.date,
```
ckiss: I don't think it's necessary to change from startTime to date.

web/public/js/controllers/dojo-event-form-controller.js
david-cahill: this sets the start time date to the start date

PR #657, opened 2015-09-21
redirect to login using $window.location.href when state.login is not available

ckiss:
ref https://github.com/CoderDojo/community-platform/issues/608


david-cahill:
lgtm :+1:

PR #660, opened 2015-09-22
small fixes to event dates

ckiss:
@david-cahill this is the fix we talked about. can you review this? thanks


david-cahill:
lgtm :+1:

PR #661, opened 2015-09-22
Install Google tag manager to handle Analytics

tangentfairy:
Will be using analytics through this in the future so we can track pledge a Dojo conversions


tangentfairy:
ref https://github.com/CoderDojo/community-platform/issues/674

ckiss:
LGTM

PR #663, opened 2015-09-22
Whitelist googletagmanager.com for external scripts

tangentfairy:
Need to add tag manager to this list

### DIFF
--- a/web/index.js
+++ b/web/index.js
@@ -121,7 +121,7 @@ server.register(Scooter, function (err) {
     mediaSrc: "'none'",
     objectSrc: "'none'",
     reflectedXss: 'block',
-    scriptSrc: "'self' 'unsafe-inline' 'unsafe-eval' https://*.googleapis.com http://www.google-analytics.com https://www.google-analytics.com https://maps.gstatic.com https://www.gstatic.com https://widget.intercom.io https://js.intercomcdn.com https://www.google.com https://apis.google.com",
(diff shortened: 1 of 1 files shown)


web/index.js
```diff
@@ -121,7 +121,7 @@ server.register(Scooter, function (err) {
     mediaSrc: "'none'",
     objectSrc: "'none'",
     reflectedXss: 'block',
-    scriptSrc: "'self' 'unsafe-inline' 'unsafe-eval' https://*.googleapis.com http://www.google-analytics.com https://www.google-analytics.com https://maps.gstatic.com https://www.gstatic.com https://widget.intercom.io https://js.intercomcdn.com https://www.google.com https://apis.google.com",
+    scriptSrc: "'self' 'unsafe-inline' 'unsafe-eval' https://*.googleapis.com http://www.google-analytics.com https://www.google-analytics.com http://www.googletagmanager.com https://www.googletagmanager.com https://maps.gstatic.com https://www.gstatic.com https://widget.intercom.io https://js.intercomcdn.com https://www.google.com https://apis.google.com",
```
david-cahill: I'd say we could remove the http url, other than that looks good :+1:

tangentfairy: We have http in for analytics for some reason... I will leave it in there for now and take it out later if it turns out we don't need it. Will probably be able to remove the analytics url altogether also.

PR #664, opened 2015-09-23
better error handling when profile uploaded avatar is not image or file is too large

ckiss:
ref https://github.com/CoderDojo/community-platform/issues/685


david-cahill:
lgtm :+1:

PR #672, opened 2015-09-24
Add titles to pages on Zen

BenChapman:
Add titles to pages on Zen, solves https://github.com/CoderDojo/community-platform/issues/671.

This is an early version, and all titles are only static. Intent is to add things like Dojo names, or User names at a future point.


david-cahill:
lgtm :+1:

PR #673, opened 2015-09-24
localizing initUserTypes on page load so they will be available as lo…

daveymoran-nearform:
…calized strings


david-cahill:
lgtm :+1:

PR #676, opened 2015-09-25
move map to pin and zoom in if only one result is found at search

ckiss:
@david-cahill this is the changed we talked about earlier.


david-cahill:
lgtm :+1:

PR #679, opened 2015-09-28
fix button size

kennanseno:
https://github.com/CoderDojo/community-platform/issues/544

buttons will now be the same height with the dropdown list/field box which is 34px.


tangentfairy:
This is fixed in the new system which is going live this week.

Also I am not sure we'd want to globally change the height of the buttons as it would mean that there would not be equal space above and below the text on the buttons. See this as an example -
34px height:
screen shot 2015-09-29 at 12 48 13

35px height:
screen shot 2015-09-29 at 12 49 17

kennanseno:
right. closing this now then. :)

PR #682, opened 2015-09-30
Tooltip on create/edit event

tangentfairy:
Add tooltip beside is public checkbox on event creation/edit page to explain what this means. Closes CoderDojo/community-platform#603


david-cahill:
lgtm :+1:

PR #689, opened 2015-10-02
Allow verified Dojos to edit their city/town

tangentfairy:
@rosalanghammer wants this enabled just for city/town


kennanseno:
lgtm

PR #693, opened 2015-10-06
fixed error after removing user from dojo

ckiss:
ref https://github.com/CoderDojo/community-platform/issues/696


tangentfairy:
:+1:

PR #695, opened 2015-10-06
fix 502 when exporting guest list

david-cahill:
ref https://github.com/CoderDojo/community-platform/issues/717

### DIFF
--- a/lib/events/cd-events.js
+++ b/lib/events/cd-events.js
@@ -35,7 +35,7 @@ module.exports = function (options) {
     var locality = (args.req$ && args.req$.cookies.NG_TRANSLATE_LANG_KEY && args.req$.cookies.NG_TRANSLATE_LANG_KEY.replace(/\"/g, '')) || args.req$ && args.req$.headers['accept-language'];
     if (!locality) locality = 'en_US';
     locality = formatLocaleCode(locality);
-    var zenHostname = args.req$.headers.host || '127.0.0.1:8000';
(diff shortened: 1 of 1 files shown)


lib/events/cd-events.js
```diff
@@ -35,7 +35,7 @@ module.exports = function (options) {
     var locality = (args.req$ && args.req$.cookies.NG_TRANSLATE_LANG_KEY && args.req$.cookies.NG_TRANSLATE_LANG_KEY.replace(/\"/g, '')) || args.req$ && args.req$.headers['accept-language'];
     if (!locality) locality = 'en_US';
     locality = formatLocaleCode(locality);
-    var zenHostname = args.req$.headers.host || '127.0.0.1:8000';
```
ckiss: @david-cahill this is a variable initialization, don't think your fix will work.

david-cahill: @ckiss how so ? It checks that `args.req$ && args.req$.headers.host` exists and if not, applies `'127.0.0.1:8000'` as the hostname

ckiss: what if `args.req$ && args.req$.headers.host ` exists? shouldn't assign something in that case too?

ckiss: I think this should be something like this now: `args.req$ && args.req$.headers.host ? args.req$.headers.host : '127.0.0.1:8000'`

ckiss:
LGTM :)

tangentfairy:
:+1:

PR #696, opened 2015-10-06
restrict users cmd_list

david-cahill:


dberesford:
if we can remove `list` lets remove it, but there are other endpoints that should require a user

PR #702, opened 2015-10-14
Fix where if a user had multiple dojos (with various permissions in each), the Manage Users link would return a 404. Fixed by passing the dojoId to the getUsersDojos service call.

dberesford:


tangentfairy:
:+1: 

Will this resolve https://github.com/CoderDojo/community-platform/issues/745 ?

dberesford:
no I don't think its related to #745

PR #705, opened 2015-10-14
Tidy up twiiter links

michaelfbradley:
Description: See issue for details. URL was broken on user profile.
Added twitter icon dojo profile

Github:
closes #695

CC: @tangentfairy

screen shot 2015-10-14 at 19 32 21

### DIFF
--- a/web/public/templates/dojos/user-profile.dust
+++ b/web/public/templates/dojos/user-profile.dust
@@ -79,7 +79,7 @@
       <div class="panel panel-default" ng-if="!hideProfileBlock('twitter') || !hideProfileBlock('linkedin')">
         <div class="panel-heading">{@i18n key="Social Media"/}</div>
         <div class="panel-body">
-          <a ng-if="profile.twitter && !hideProfileBlock('twitter')" href="https://{{ profile.twitter }}"><i class="fa fa-twitter fa-2x col-lg-3"></i></a>
+          <a ng-if="profile.twitter && !hideProfileBlock('twitter')" href="https://twitter.com/{{ profile.twitter }}"><i class="fa fa-twitter fa-2x col-lg-3"></i>{{ profile.twitter }}</a>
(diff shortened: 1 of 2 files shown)


web/public/templates/dojos/user-profile.dust
```diff
@@ -79,7 +79,7 @@
       <div class="panel panel-default" ng-if="!hideProfileBlock('twitter') || !hideProfileBlock('linkedin')">
         <div class="panel-heading">{@i18n key="Social Media"/}</div>
         <div class="panel-body">
-          <a ng-if="profile.twitter && !hideProfileBlock('twitter')" href="https://{{ profile.twitter }}"><i class="fa fa-twitter fa-2x col-lg-3"></i></a>
+          <a ng-if="profile.twitter && !hideProfileBlock('twitter')" href="https://twitter.com/{{ profile.twitter }}"><i class="fa fa-twitter fa-2x col-lg-3"></i>{{ profile.twitter }}</a>
```
tangentfairy: The issue with this is that the placeholder on this page https://github.com/CoderDojo/cp-zen-platform/blob/cf2e0c2eaecd208b91fc1bba1e2871e8f155059b/web/public/templates/profiles/general-info.dust#L90 indicates that you need to enter the "twitter.com" part of the URL first. So if I enter "twitter.com/CoderDojo" here now I will end up with "twitter.com/twitter.com/CoderDojo" here. 

I think we need to go with the full solution discussed here https://github.com/CoderDojo/community-platform/issues/695 where we show the twitter.com part of the URL so it is explicit that you need to only enter your username. We'll need to update the placeholder also!

michaelfbradley: Cool - I'll look into validating this at the point of input, as we want the twitter handle to be saved in a consistent format, so pretty much as suggested above.

I'll sort within the next 24hrs :metal:

tangentfairy: No rush - make sure you sign up for 24pullrequests.com when contributing to open source projects - and you don't need to put in one PR per day - just 24 by Christmas. You can earn digital badges :)

tangentfairy:
Going to close this until it is ready for a re-review! Thank you!

PR #706, opened 2015-10-14
Add message to top of past event to explain it is not editable

tangentfairy:
Show a message at the top of the edit event page for past events


dberesford:
LGTM

PR #707, opened 2015-10-14
fix blue button height

kennanseno:
https://github.com/CoderDojo/community-platform/issues/544

fixed blue button height so it's similar to the 2 other types(default, success) of button.


tangentfairy:
:+1: Looks good!

PR #716, opened 2015-10-17
#746 Added empty events warning

Arrayindex:
Added check for empty events array and added new text to locale


tangentfairy:
:+1: Excellent!

PR #724, opened 2015-10-23
Add message to My Dojos page when no dojo in list

Aryess:
Close #520

### DIFF
--- a/web/locale/en_US/messages.po
+++ b/web/locale/en_US/messages.po
@@ -2642,3 +2642,6 @@ msgstr "suggest features"
 
 msgid "to help make it better."
 msgstr "to help make it better."
+
+msgid "No dojo"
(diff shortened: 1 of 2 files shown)


web/locale/en_US/messages.po
```diff
@@ -2642,3 +2642,6 @@ msgstr "suggest features"
 
 msgid "to help make it better."
 msgstr "to help make it better."
+
+msgid "No dojo"
+msgstr "You don't have any Dojos - click here to start a dojo!"
```
tangentfairy: The msgid string is always the same as the msgstr in the English file. The msgid will be English in every po file and the msgstr is the corresponding translation (which is just the same for English in this case).

tangentfairy:
:+1:

PR #725, opened 2015-10-30
Swagger API descriptions

mihaidma:
Added swagger API descriptions

On many of them put `required()` constraints and even detailed parameter constraints. Adapted and verified the constraints not to affect the current functionality by allowing current values. Example: for **verifiedAt**: put the constraint `Joi.date()` but also allowed the alternate value of **null** (`Joi.string().valid(null))`)

At the beggining of **dojos.js** file made a **joiValidator** where gathered functionality that can be reused. In time when adding Swagger also on the other APIs this functionality should evolve into a library/module used by all APIs.

It can be enhanced by adding more restrictions. Example: not allowing **null** values when a **guid** is expected or putting `required()` on more fields to ensure that all API calls respect a certain format.

### DIFF
--- a/lib/dojos.js
+++ b/lib/dojos.js
@@ -32,7 +32,7 @@ var joiValidator = {
   country: function() {
     return Joi.object().keys({
       countryName: Joi.string().required(),
-      countryNumber: Joi.string().required(),
+      countryNumber: Joi.number().integer().required(),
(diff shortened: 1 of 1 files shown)


dberesford:
LGTM - squashed and merged: 55e23e1bd86aa0b220daf242d078183ba3027ff7

PR #727, opened 2015-11-06
Enable Intercom for all accounts except ninjas/youths

tangentfairy:
Enable intercom for parents and mentors. Resolves https://github.com/CoderDojo/community-platform/issues/780

### DIFF
--- a/web/public/js/services/intercom-service.js
+++ b/web/public/js/services/intercom-service.js
@@ -8,6 +8,11 @@ angular.module('cpZenPlatform').factory('intercomService', function ($localStora
     return _.contains(user.roles, 'cdf-admin');
   }
 
+  function userIsNinja(user) {
+    var initUserTypeStr = JSON.parse(user.initUserType);
+    return initUserTypeStr.name === 'attendee-o13' || initUserTypeStr.name === 'attendee-u13';
+  }
+
   function userIsChampion(user, cb) {
     cdUsersService.isChampion(user.id, cb,
       function (err) {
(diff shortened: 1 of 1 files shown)


web/public/js/services/intercom-service.js
```diff
@@ -62,12 +69,8 @@ angular.module('cpZenPlatform').factory('intercomService', function ($localStora
       if (user) {
         if (userIsCDFAdmin(user)) {
```
ckiss: @tangentfairy from what I can see I think you can remove this `if (userIsCDFAdmin(user)) {` and leave only this 
`if(!userIsNinja(user)){`. Otherwise looks good :)

web/public/js/services/intercom-service.js
```diff
@@ -8,6 +8,11 @@ angular.module('cpZenPlatform').factory('intercomService', function ($localStora
     return _.contains(user.roles, 'cdf-admin');
   }
 
+  function userIsNinja(user) {
+    var initUserTypeStr = JSON.parse(user.initUserType);
```
ckiss: this is initialized outside function scope too. I don't think you need to do it twice

tangentfairy: It is initialized inside [bootIntercom()](https://github.com/CoderDojo/cp-zen-platform/blob/enable-intercom-for-all-but-ninjas/web/public/js/services/intercom-service.js#L37), not outside function scope

ckiss: Ah yes, you're right, but either way, can be initialized only once at the top. lint shouldn't let this go through I think :)

tangentfairy: Added that now.

PR #729, opened 2015-11-06
swagger validation enhancements

mihaidma:
Added `joiValidator.guid()` that allows both `guid()` and `string()`
Allow `object` or `null` for `country, state, city`


mihaidma:
It fixes the #782

tangentfairy:
:+1:

PR #730, opened 2015-11-08
fix #547 from the community-platform repo

ferry-:
changed the event description input field to a html input box like the one used in the dojo form.
changed the event template to display the description as html

### DIFF
--- a/web/public/templates/dojos/events/apply.dust
+++ b/web/public/templates/dojos/events/apply.dust
@@ -18,7 +18,7 @@
       <p>{{ event.city.nameWithHierarchy }}</p>
     </div>
     <div class="col-md-6">
-      <p>{{ event.description }}</p>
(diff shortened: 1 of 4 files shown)


web/public/templates/dojos/events/apply.dust
```diff
@@ -18,7 +18,7 @@
       <p>{{ event.city.nameWithHierarchy }}</p>
     </div>
     <div class="col-md-6">
-      <p>{{ event.description }}</p>
+      <p><div ng-bind-html="event.description"></div></p>
```
tangentfairy: We don't need to add an extra `div` element here, we can just do `<p ng-bind-html="event.decsription">`. Unless you really need it to be a block element (in which case you could have changed the `p` to a `div`.

tangentfairy:
:+1:

PR #732, opened 2015-11-09
save dojo lead swagger enhancements

mihaidma:
The 2 fields were optional but in case they exist allow empty strings for `Technical Experience` and `Experience working with youth` fields


tangentfairy:
:+1:

PR #734, opened 2015-11-09
api validation enhancements

mihaidma:


lib/dojos.js
```diff
@@ -126,7 +127,7 @@ var joiValidator = {
       mysqlUserId: Joi.any(),
       firstName: Joi.any(),
       lastName: Joi.any(),
-      roles: Joi.array().required(),
+      roles: Joi.array(),
```
tangentfairy: I think a role will be required because you can either be {basic-user} or {cdf-admin}

PR #740, opened 2015-11-12
Allow empty values for times, twitter and website fields

tangentfairy:
People cannot edit their Dojo listing because the validation is still too strict. Fixes https://github.com/CoderDojo/community-platform/issues/793


dberesford:
LGTM

PR #742, opened 2015-11-13
removed unneeded calls to profiles/user-profile-data

ckiss:
This is just a cleanup and fixes part of https://github.com/CoderDojo/community-platform/issues/745 specifically Mihai's use case where he switches from My Dojos to Dojo Listing a few times


tangentfairy:
:+1: After testing this is an improvement at least!

PR #743, opened 2015-11-17
Fix issue with erasing notes, email, time, website and twitter fields

tangentfairy:
Champions not able to edit their Dojo listing because validation is too strict, these are all optional fields

ref https://github.com/CoderDojo/community-platform/issues/793


mihaidma:
:+1: LGTM

PR #744, opened 2015-11-17
Fix validation on optional text fields on setup your dojo step

tangentfairy:
Users cannot pass this step due to validation being too strict...


mihaidma:
:+1:

PR #754, opened 2015-11-25
redirect to referer if exists after editing profile

ckiss:
ref https://github.com/CoderDojo/community-platform/issues/805


tangentfairy:
Tested and looks good :+1:

PR #756, opened 2015-11-25
show error when child profile doesn't get saved

ckiss:
ref https://github.com/CoderDojo/community-platform/issues/801

@tangentfairy error is shown now


tangentfairy:
:+1:

PR #762, opened 2015-11-30
Add banner to top of profile

tangentfairy:
Two banners - one to explain that you are not booked in to an event once you fill out your profile.

The other to explain that not all fields on your profile are public.

It seems like users think that once they have signed up to the system they are booked into an event which is not so.

Also add a link in the events explanation to the ninjas section of the profile


mihaidma:
:+1:

PR #763, opened 2015-11-30
Updates to registration page

tangentfairy:
- Add message to registration page to explain that the user is registering an account rather than registering for an event
- Change "register" to "register an account" to make it explicit what the registration page does

Ref https://github.com/CoderDojo/community-platform/issues/816


mihaidma:
:+1:

PR #768, opened 2015-12-05
Passing user creation date to Intercom. Fix for #827

josmas:
Fix for ref https://github.com/CoderDojo/community-platform/issues/827.

Do you guys have an Intercom test site I could push this against for testing? cc/ @tangentfairy


tangentfairy:
The development mode for Intercom needs to be set up (as per https://github.com/CoderDojo/community-platform/issues/699)

This looks good to me, so :+1: and I can confirm this is working after deployment.

PR #772, opened 2015-12-09
I 834 move add ninja links

tangentfairy:
- Rename ninjas to youths in various areas where it was confusing for parents and champions
- Add additional links to create U13 and O13 ninjas to the view profile page

### DIFF
--- a/web/public/js/controllers/manage-event-applications-controller.js
+++ b/web/public/js/controllers/manage-event-applications-controller.js
@@ -57,8 +57,9 @@
       }
     }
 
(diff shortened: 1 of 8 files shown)


web/public/js/controllers/manage-event-applications-controller.js
```diff
@@ -58,7 +58,7 @@
     }
 
     $scope.ticketTypes = [
-      {name: 'ninja', title: 'Ninja' },
+      {name: 'ninja', title: 'Youth' },
```
josmas: consistency of name change here; same as in https://github.com/CoderDojo/cp-events-service/pull/90

PR #775, opened 2015-12-13
Filters inactive dojos in list and deletes an unused controller. Fix for #824

josmas:
Fix for https://github.com/CoderDojo/community-platform/issues/824

The fix is far from ideal but it is the same ***hack*** used for the maps view. As noted in https://github.com/CoderDojo/community-platform/issues/831, refactoring this code would be great, but it turns out it's not as simple as expected. The reason is that the *stage* field is an integer 0 to 4. It works fine for filtering to one particular stage (sending a query like *stage = 1*) but we want something like *stage !== 4* and there's no way to send that over in a query parameter. There are a couple of routes that use custom SQL in the service and hardcode that value. That could be a solution, but it means creating a new route for something that a *query* parameter should be able to do.

Also, the list controller is not used anymore, all the code is in the map controller (it took a while to figure that out, so deleting unused file here too so there's no confusion in the future).


tangentfairy:
This is a handy short term solution until the Seneca team can provide a better way to filter the query. We have passed on issue https://github.com/CoderDojo/community-platform/issues/831 to them :+1:

mirceaalexandru:
postgres version 0.8.3 now supports:

ne$: .list$({ f1: {ne$: v1} }) for not-equal.
eq$: .list$({ f1: {eq$: v1} }) for equal.
lte$: .list$({ f1: {lte$: 5} }) for less than or equal.
lt$: .list$({ f1: {lt$: 5} }) for less than.
gte$: .list$({ f1: {gte$: 5} }) for greater than or equal.
gt$: .list$({ f1: {gt$: 5} }) for greater than.
in$: .list$({ f1: {in$: [10, 20]} }) for in. in$ operator accepts only values of type array.
nin$: .list$({ f1: {nin$: ['v1', 'v2']} }) for not-in. nin$ operator accepts only values of type array.

Note: you can use any operators described above together.

Another change was that sort$ was fixed. To be consistent with the other seneca store the sort order was reversed.

josmas:
Hi @mirceaalexandru thanks for the info and apologies in advanced for the ***wide*** question, I am only starting to get familiar with Seneca: Can that be leveraged from the client? as in, passing the operators in a *query*? Or does it have to be changed in the service itself?

mirceaalexandru:
postgres store version bump to 1.0.0.

@josmas I really think that this should be fixed on service itself but I don't have knowledge about actual implementation of this project. 

However if you are displaying sorted/paginated table data (as an example) then you might be forced to pass query filters/sorting from client as part of request.

tangentfairy:
Going to merge this and we'll fix #831 in the long term

PR #778, opened 2015-12-17
Filtering invites by ticket type. Fix for #790

josmas:
Only the appropriate *roles* should appear in the drop downs now. There is quite a bit of repetition between two controllers, but one seems to be basically copy and paste from the other one, even though the service calls are done in different places. Bit convoluted but couldn't do much more as it stands... please review!

### DIFF
--- a/web/public/js/controllers/apply-for-event-controller.js
+++ b/web/public/js/controllers/apply-for-event-controller.js
@@ -11,6 +11,27 @@
       } else {
         $scope.showEventInfo(eventIndex, $scope.event.id);
       }
+    };
+
+    // Functions to map Service ticket types to front-end ticket types.
+    // Front end types are: 1.ninja, 2.guardian-parent 3.mentor, 4.other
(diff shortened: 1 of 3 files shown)


web/public/js/controllers/apply-for-event-controller.js
```diff
@@ -11,6 +11,27 @@
       } else {
         $scope.showEventInfo(eventIndex, $scope.event.id);
       }
+    };
+
+    // Functions to map Service ticket types to front-end ticket types.
+    // Front end types are: 1.ninja, 2.guardian-parent 3.mentor, 4.other
+    function isNinja(user) {
```
tangentfairy: These functions seem like the kind of thing that might be used in a lot of places - [would this file](https://github.com/CoderDojo/cp-zen-platform/blob/master/web/public/js/controllers/header-controller.js) be a good place for something like that?

tangentfairy:
Looks good! :+1:

PR #779, opened 2015-12-18
Fix for sorting crashes when managing dojos. #848

josmas:
As per title. It was an issue with not allowing certain fields in the query. Fix for https://github.com/CoderDojo/community-platform/issues/848


tangentfairy:
Looks good! :+1:

PR #780, opened 2015-12-18
changes in ordering due to upgrade of seneca-postgresql-store

josmas:
This is 1 in 4 pull requests for https://github.com/CoderDojo/community-platform/issues/839

(3 services + 1 platform)

### DIFF
--- a/lib/dojos.js
+++ b/lib/dojos.js
@@ -279,7 +279,11 @@ exports.register = function (server, options, next) {
           limit$: Joi.number().integer().min(0).optional(),
           skip$: Joi.number().integer().min(0).optional(),
           sort$: Joi.object().keys({
-            created: Joi.number().valid(-1).valid(1).optional()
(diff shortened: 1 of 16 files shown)


tangentfairy:
Changing the sort in manage users or manage dojos gives me a bad request on this branch, but is okay on Master. Did an npm install also. Am I missing something?

josmas:
@tangentfairy 
Testing this is a bit tricky, sorry, I should have added instructions.

Before npm install you need to delete the previous version of the store. Don't know of an automated way of doing this; I just go into node_modules and `rm -rf` the modules I don't want; then `npm install` should bring the new one with all new dependencies. Does that make sense? Otherwise the older module might be found first and loaded (guessing here).

tangentfairy:
Got it! Will give that a whirl, thanks!

tangentfairy:
Looks good! :+1:

PR #782, opened 2015-12-19
fix for utf-8 issues with translations

josmas:
This was a tricky one, because it only happens while the app is loading (anything translated within a template will be properly handled by the encoding in the main template, but this happens on app initialisation). It seems to be an issue with the library itself as explained [here](https://github.com/angular-translate/angular-translate/issues/1101#issuecomment-119300028). Thoughts?

### DIFF
--- a/web/public/js/init-dashboard.js
+++ b/web/public/js/init-dashboard.js
@@ -430,8 +430,8 @@
       function($translateProvider) {
         $translateProvider.useUrlLoader('/locale/data?format=mf')
         .useCookieStorage()
-        .useSanitizeValueStrategy('sanitize')
-        .registerAvailableLanguageKeys(['en_US', 'it_IT'])
+        .useSanitizeValueStrategy('sanitizeParameters')
+        .registerAvailableLanguageKeys(['en_US', 'it_IT', 'de_DE', 'pt_PT'])
         .uniformLanguageTag('java')
         .determinePreferredLanguage()
         .fallbackLanguage('en_US');


josmas:
There's phantomJS failure in circleCI; it's happened before; all test run fine locally.

josmas:
Also, forgot to mention, this is a fix for https://github.com/CoderDojo/community-platform/issues/815

niccokunzmann:
Should we re-run the tests and it works then?

josmas:
@niccokunzmann You mean restart the tests in circleCI? That should make the checks pass, unless I forgot something, but it works fine locally, and circleCI gets stuck at times.

For the fixes to appear on the site though, this PR will have to be reviewed, and if all is good, merged and then deployed to the production site. I believe Ursula is away for a few days, so this won't be happening probably until the new year.

Does that answer your question?

tangentfairy:
This looks good - I reran the tests! :+1:

tangentfairy:
This works when logged in, but not when _not_ logged in. I think that's because the file you added it to must only be loaded on login?

josmas:
That's odd. Will have a look. Might be the way the file is written (encoding)... Are the language files (other than English) grabbed automatically from the translation server or downloaded manually during a production deploy?

tangentfairy:
Downloaded during deploy. I don't think that is the issue because we still use the same translation file when you are not logged in.

josmas:
mmmm... may be unlikely but worth checking; you have ssh access to the different folders, right? Could you run a `file` command on the translation files? Just to make sure they are ***UTF-8***; something like:

    $ file messages.po
    messages.po: GNU gettext message catalogue, UTF-8 Unicode text, with very long lines

tangentfairy:
@josmas this one slipped my mind as it was a closed PR! Here is the output on product, it seems to be the same:

```
$ file messages.po
messages.po: GNU gettext message catalogue, UTF-8 Unicode text, with very long lines
```

josmas:
@tangentfairy Yup, file seems to be good. I'll have a look during the hacking session this weekend!

tangentfairy:
Great, thanks! :)

PR #788, opened 2016-01-06
fix event attendee list on mobile

kennanseno:
https://github.com/CoderDojo/community-platform/issues/861


kennanseno:
not sure why the check failed, works when I test it. @tangentfairy

PR #793, opened 2016-01-08
Update map API key

tangentfairy:
I'm not sure if Google nuked the last key or what happened, but update it anyway...


niccokunzmann:
I see two files changed and do not think that this change can make anything worse.

niccokunzmann:
However, I do not understand how the error could relate to this.

niccokunzmann:
When can I expect this to be deployed?

PR #798, opened 2016-01-15
Added we need mentors and booking dialog beneath contact section

adhorrig:
Moved a couple of sections as per Rosa and Nicco's request


tangentfairy:
:+1: Looks good!

PR #799, opened 2016-01-19
Fix Issue#562

flakyfilibuster:
I didn't have to make another call to the service after all as we're loading the dojos info on page load anyways.
I cache the address and city for later use and grab it for that variable when the checkbox is checked.

Please let me know if that's a viable approach or if I should change something.

Thanks

### DIFF
--- a/web/public/templates/dojos/events/dojo-event-form.dust
+++ b/web/public/templates/dojos/events/dojo-event-form.dust
@@ -19,8 +19,7 @@
       <span tooltip="{@i18n key="Public events are viewable by anyone visiting your Dojo listing. Private events can only be seen by members of your Dojo."/}" tooltip-placement="right" class="info-button fa fa-info-circle event-tooltip"></span>
     </div>
 
-    <div class="form-group">
-    <div class="form-group form-inline">
+    <div class="form-group form-inline" ng-show="!eventInfo.status">
       <div class="checkbox">
         <input type="checkbox" id="prefill-address" ng-model="eventInfo.prefillAddress" ng-click="prefillDojoAddress()">
         <label for="prefill-address">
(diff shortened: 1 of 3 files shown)


web/public/templates/dojos/events/dojo-event-form.dust
```diff
@@ -20,6 +20,16 @@
     </div>
 
     <div class="form-group">
+    <div class="form-group form-inline">
+      <div class="checkbox">
+        <input type="checkbox" id="prefill-address" ng-model="eventInfo.prefillAddress" ng-click="prefillDojoAddress()">
+        <label for="prefill-address">
+          {@i18n key="Use Dojo listing address"/}
```
tangentfairy: This needs to be added to messages.po also :)

flakyfilibuster:
done @tangentfairy

tangentfairy:
This has some funny behaviour in edit mode - it works fine on a complete new event:

- On a published or draft event, if I tick the box it appears to erase the address.
- There is a similar issue if you enter details on a draft, and instead of clicking "save draft", refresh the page. The form doesn't remember what address was entered when you use the checkbox, but it does if you manually enter the address. I think this is because we use local storage on any uncompleted forms, but it isn't triggered when you use the checkbox.
- Can we hide this checkbox if it is a published event? At that stage the address needs to be added anyway.

We'll just have to sort out the issue with toggling the address on a draft event then.

flakyfilibuster:
Hey, yeah currently the behavior is set to either add the dojos listed address or remove the address in the city & address fields when unchecked.

I'll save the address & city to the localstorage as soon as it's checked. That way it'll be remembered on a reload.

Let me know how you want it to behave when an address is entered and the user unchecks the option.


Cheers

flakyfilibuster:
Have a look @tangentfairy

tangentfairy:
This all looks great now! Will deploy in the morning :+1:

PR #781, opened 2015-12-18
Fix for crashing sort by email in manage users list

josmas:
No issue open as such for this one. To reproduce:

  - As a manager click **My Dojos**
  - Click on **Users - Manage**
  - Sort by *name* works, but sort by ***email*** crashes as in https://github.com/CoderDojo/community-platform/issues/848

PR #800, opened 2016-01-22
Fix incorrect total waitlist number when adding new attendants. #781

machineperson:
https://github.com/CoderDojo/community-platform/issues/781

Don't decrement the number of people on the waitlist when the waitlist is empty.


tangentfairy:
:+1: Looks perfect!

PR #801, opened 2016-01-23
Clean up the 404 pages to use fewer unnecessary divs…

jyujin:
… and replace the a/br sequence with an unordered list (as it is one).

It's a pet peeve of mine, I know :). It should render the same as the previous iteration...


tangentfairy:
Looks good! :+1:

PR #802, opened 2016-01-23
Fix the charter-page link not working correctly on the 404 page when logged in.

jyujin:
It's set up in the dashboard controller, which is not pulled in for the 404, but setting it up in the base state still works. @machineperson actually figured this one out.

This is for issue https://github.com/CoderDojo/community-platform/issues/866


tangentfairy:
Looks good! :+1:

PR #804, opened 2016-01-23
feature/883-filter-applicants

Wardormeur:
PR for CoderDojo/community-platform#883


tangentfairy:
As we discussed:
- We could make the arrow beside applicant name the same as the style on manage dojos
- We can add the same sorting function to the application date, so that by default it is still sorted by application date
- There seems to be a weird issue when sorting by application name: a name beginning with lowercase shows up as last, even though it begins with a "k"! Let me know if this is actually an issue with the plugin we are using and I can report it to the team that maintain it.

Wardormeur:
Added !

tangentfairy:
Looks great! :+1:

PR #749, opened 2015-11-23
user API fixes and cleanup

ckiss:
Did some refactoring of the users API:
-removed unused API endpoints (load, update)
-added endpoint to load previous founder(protected, user needs to be logged in)
-fixed handler by adding lodash-deep dependecy
-some cleanup & other small fixes

PR #805, opened 2016-01-23
Fixed tooltip size

gytdau:
Enforces a width of 30em for tooltips in the 'Setup your Dojo' form.
This fixes the tooltips being too tall and difficult to read.

Closes #749


tangentfairy:
Looks much better, cheers!

PR #806, opened 2016-01-23
Fixed 'y' in search bar for Firefox users

gytdau:
This might win the prize for the most inconsequential commit ever pushed.

The 'y' is clipped for Firefox users. It's not clipped anymore.


tangentfairy:
Excellent! Nice to get these niggly issues boxed off :+1:

PR #807, opened 2016-01-23
#881 - Delete ticket application error

andresiglesias:
Solves issue [#881](https://github.com/CoderDojo/community-platform/issues/881)

### DIFF
--- a/web/public/js/controllers/manage-event-applications-controller.js
+++ b/web/public/js/controllers/manage-event-applications-controller.js
@@ -322,7 +322,7 @@
       function updateDeleted() {
         application.updateAction = 'delete';
         application.deleted = true;
-        successMessage = $translate.instant('Ticket for') + ' ' + application.name + ' ' + $translate.instant('has been successfully deleted');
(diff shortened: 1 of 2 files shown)


web/public/js/controllers/manage-event-applications-controller.js
```diff
@@ -322,6 +322,7 @@
       function updateDeleted() {
         application.updateAction = 'delete';
         application.deleted = true;
+        successMessage = $translate.instant('Ticket for') + ' ' + application.name + ' ' + $translate.instant('has been successfully deleted');
```
tangentfairy: I have just realised it might be better if we say it like:
Ticket for the following application has been successfully deleted: application.name"
Because otherwise this will be put up to the translation service in separate chunks.

Also, we will need to add the string [to the messages.po file here](https://github.com/CoderDojo/cp-zen-platform/blob/master/web/locale/en_US/messages.po).

tangentfairy:
:+1: Looks good!

PR #810, opened 2016-01-23
Solve issue #846 allowing public access to controller dojo-detail.

yannart:
Add the controller 'dojo-detail' to the list of public states to avoid redirecting to the profile form.


tangentfairy:
Excellent! Works perfectly :+1:

PR #816, opened 2016-02-11
Fix for issue with recurring events in certain timezones

adhorrig:
Issue 893:

When creating a recurring event using the dates:
Every Sunday from the 14th February -> 6th March.
The dates would change once you save the draft and go back to edit it, the dates would no longer display as the 14th February -> 6th March. They would be changed to the 15th February -> 7th March.

Video : https://drive.google.com/drive/folders/0B1eLynE0aU3mbVRmYmtlOHdtcjg

### DIFF
--- a/web/public/js/controllers/dojo-event-form-controller.js
+++ b/web/public/js/controllers/dojo-event-form-controller.js
@@ -3,7 +3,7 @@
   'use strict';
 
   function getEveryTargetWeekdayInDateRange(startDateTime, endDateTime, targetWeekday, eventType) {
-    endDateTime = moment.utc(endDateTime).add(1, 'days');
+    endDateTime = moment.utc(endDateTime).add(1, 'days'); //Fixes the while loop on #28 exiting a day short
     var currentDate = startDateTime;
     var dates = [];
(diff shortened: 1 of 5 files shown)


tangentfairy:
@ckiss I know you had an opinion on how we did this before, maybe you could take a look and let me know what you think please? :)

tangentfairy:
Related to https://github.com/CoderDojo/community-platform/issues/893

web/public/js/controllers/dojo-event-form-controller.js
```diff
@@ -3,6 +3,7 @@
   'use strict';
 
   function getEveryTargetWeekdayInDateRange(startDateTime, endDateTime, targetWeekday, eventType) {
+    endDateTime = moment.utc(endDateTime).add(1, 'days');
```
tangentfairy: Can you leave a comment explaining why we are adding a magic time of one day here?

adhorrig:
When we do currentDate <= endDate, the while loop exists a day short

tangentfairy:
:+1:

PR #818, opened 2016-02-16
Issue#760

flakyfilibuster:
@tangentfairy: fix for https://github.com/CoderDojo/community-platform/issues/760


web/public/js/controllers/manage-dojo-users-controller.js
```diff
@@ -69,11 +69,13 @@ function cdManageDojoUsersCtrl($scope, $state, $q, cdDojoService, alertService,
       getUsersDojosLink,
       loadDojoUsers,
       retrieveEventsAttended,
-      getTotalCount
+      getTotalCount,
+      usSpinnerService.stop('manage-dojo-users-spinner');
```
Wardormeur: This needed to be encapsulated into a function, it was probably working for you because it was falling into the error callback by the absence of badge system :)
Elsewhat, :+1:

PR #823, opened 2016-02-23
Preselect user type when join dojo is selected

tjama:
This fixes the following issue: https://github.com/CoderDojo/community-platform/issues/903

Extra parameter is added when changing the route to registration and then used to pre-select proper user type.


Wardormeur:
LGTM :+1:

PR #824, opened 2016-02-23
dojo-list-index-controller -> dojo-events-list-controller

denniskenny:
This file appears to have been renamed in a previous refactor, updating dependencies to point at the new name.


ArayB:
Closing as this is very old.

PR #832, opened 2016-03-09
I736 dojo listing capacity field

adhorrig:
Ref: https://github.com/CoderDojo/community-platform/issues/736


lib/dojos.js
```diff
@@ -323,6 +323,7 @@ exports.register = function (server, options, next) {
             alpha3: joiValidator.alpha3().required(),
             address1: Joi.string().required(),
             coordinates: Joi.string().required(),
+            attendee: Joi.string().required(),
```
Wardormeur: I would have preferred something as expectedAttendees to avoid misunderstanding upon the goal of this attribute.
When i read the code, i thought it was required to have an attendee as a string when creating a dojo, sounded strange :p
Also, shoudn't validation be a number instead of a string?

web/public/js/controllers/start-dojo-wizard-controller.js
```diff
@@ -647,6 +647,7 @@ function startDojoWizardCtrl($scope, $window, $state, $location, auth, alertServ
       var lsdl = $localStorage[currentUser.id].dojoLead.dojoListing;
 
       if(lsdl.name) $scope.dojo.name = lsdl.name;
+      if(lsdl.attendee) $scope.dojo.attendee = lsdl.attendee;
```
Wardormeur: If it's supposed to be a number, do the type change here (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number)

web/public/templates/dojos/dojo-form.dust
```diff
@@ -147,6 +147,20 @@
       </div>
 
       <div class="row cd-input-row" ng-enter="noop()">
+          <div class="form-group">
+            <label for="dojoAttendee" class="col-lg-4 cd-form-label1">
+              <span>{@i18n key="Dojo Attendees"/}*</span>
+            </label>
+            <div class="col-lg-8">
+              <input type="text" class="form-control" form="createDojoForm" required required-message="'{@i18n key="Dojo Attendees is empty"/}'" ng-model="dojo.attendee" name="dojoAttendee" id="dojoAttendee" placeholder="{@i18n key="e.g. 10-15"/}" ng-change="updateLocalStorage('dojoListing','attendee', dojo.attendee)">
```
Wardormeur: Or if it's a number, change the type of the input to number?

lib/dojos.js
adhorrig: I agree - it could be more clear. I've refactored to make it more understandable and to use input type = number also.
Thanks! :)

tangentfairy:
Looks good but don't forget this will need to be added to the Salesforce integration at some stage.

PR #834, opened 2016-03-10
I789 dojo frequency

adhorrig:
Ref: https://github.com/CoderDojo/community-platform/issues/789

Splits the "times" filed into frequency and time. Also makes the time field required, as it wasn't before.


adhorrig:
closing - see: https://github.com/CoderDojo/cp-zen-platform/pull/783

PR #836, opened 2016-03-11
Replace Ninja with User

adhorrig:
Makes the wording less off-putting 

Ref : https://github.com/CoderDojo/community-platform/issues/917#issuecomment-195310043
https://github.com/CoderDojo/community-platform/issues/934#issuecomment-195301232

### DIFF
--- a/web/locale/en_US/messages.po
+++ b/web/locale/en_US/messages.po
@@ -1597,8 +1597,8 @@ msgstr "Select a user type to invite..."
 msgid "Search by name"
 msgstr "Search by name"
 
-msgid "Are you sure you want to award this badge to the below Ninja?"
-msgstr "Are you sure you want to award this badge to the below Ninja?"
+msgid "Are you sure you want to award this badge to the below user?"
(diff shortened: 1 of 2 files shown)


niccokunzmann:
@adhorrig what is the process to merge this? If I look aver it and tested it and say OK. Can you merge it? What needs to be done to merge something?

adhorrig:
@niccokunzmann I don't have merge access myself so I will not be able to do any merging. General process for merging is that we confirm that the code change fixes the issue, and then the PR is merged :)

Wardormeur:
@niccokunzmann I just asnwered on gitter, basically we're in a transition phase for the development team, that's why nothing is actually merge/deployed
Sorry, should have told it earlier :|

tangentfairy:
:+1: from me

Wardormeur:
LGTM :+1: 
However, is there a reason to not use variables in the translation (instead of "below user", replace by the targeted {{username}})? Is it to ease the Crowdin translation? 
(ref https://bitbucket.org/manolakis/dustjs-i18n)

PR #839, opened 2016-03-20
Fixes translation error from #934

adhorrig:
Ref: https://github.com/CoderDojo/community-platform/issues/934

Specific: "Join Request Sent!" was not translated


tangentfairy:
Looks good :+1:

PR #841, opened 2016-03-22
Add support for existing crowding languages but not dynamiccaly added

Wardormeur:
Again, care about this double init file


adhorrig:
Tested and looks good :+1: .. Using a language with non-english letters also makes it easier to see which translations are missing

PR #843, opened 2016-03-24
822 - Remove required email for o13 (when registred through a parent)

Wardormeur:
Partial workaround for 822, since o13 won't be able to login until parent add them an email address
#https://github.com/CoderDojo/community-platform/issues/822


adhorrig:
Tested :+1: from me

PR #844, opened 2016-03-24
Adding information for unsupported urls languages

Wardormeur:
Relative to https://github.com/CoderDojo/cp-dojos-service/pull/234
Aslo fix City/town not being translated

### DIFF
--- a/web/public/templates/dojos/dojo-detail.dust
+++ b/web/public/templates/dojos/dojo-detail.dust
@@ -13,7 +13,7 @@
             <strong>{@i18n key="Private Dojo"/}:</strong> {@i18n key="This means this Dojo is only open to specific people (for example, students in a school). Please do not contact this Dojo."/}
           </div>
           <div class="alert alert-warning text-center" ng-if="dojo.stage.value === 4" role="alert">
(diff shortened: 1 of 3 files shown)


web/public/templates/dojos/dojo-detail.dust
```diff
@@ -27,7 +27,7 @@
                 <td ng-if="dojo.address1" ng-bind-html="dojo.address1">, <span ng-if="dojo.address2" ng-bind-html="dojo.address2"></span></td>
               </tr>
               <tr>
-                <td class="col-xs-2"><b>{@i18n key="City/Town"/}</b></td>
+                <td class="col-xs-2"><b>{@i18n key="CityTown"/}</b></td>
```
tangentfairy: Ideally you'd just make the key [on this line](https://github.com/CoderDojo/cp-zen-platform/pull/844/files#diff-5bcf96d3f271bdd27ce64c408b7b5972L438) the same as the value of `City/Town`. You can escape slashes by adding a `\` in front of slashes if you need to. The way the messages.po file works is that the key is always the same as the value in the English version of the file, and then the English key is used in the foreign versions.

Wardormeur: I didn't change the key because of the existing translations, but yeah

PR #845, opened 2016-03-24
Event page re-design

adhorrig:
Event page split into two panels 

cc @Giustina @rosalanghammer

### DIFF
--- a/web/public/css/base/_form.less
+++ b/web/public/css/base/_form.less
@@ -24,3 +24,16 @@ textarea {
     border: 1px solid @error-color !important;
   }
 }
+
+input[type=number]{
+    width: 60px;
+}
+
+legend::after{
+  width: 50% !important;
+}
+
+.padding { margin-left: 10px; }
+@media (max-width: 1200px) {
+   .padding { margin-left: 0px; }
+}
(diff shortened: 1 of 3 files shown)


tangentfairy:
The "Book" button is cut off on desktop view for me. Latest version of Firefox.

https://zen.coderdojo.com/dashboard/dojo/ie/cork/cork/nsc-mahon-cork

tangentfairy:
Ideally when you are making commits it would be great if you could squash all commits into one before doing a pull request.

This means [when we look at the history here for master](https://github.com/CoderDojo/cp-zen-platform/commits/master), it is easy to see your commit labelled "Event page re-design". At the moment we see many commits including one that says "more layout changes" which is not descriptive so we do not know what the objective of the commit was.

You could [follow this simple way](http://stackoverflow.com/questions/5308816/how-to-use-git-merge-squash) of squashing your commits before putting a pull request into CoderDojo:Master.

[Here's a fun article about why to squash commits and why it is popular on open source repositories.](https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history)

Cleaner commits = more contributors = happy contributors = makes me cry tears of joy :star2:

adhorrig:
Thanks for the information and resources! I'll definitely be doing this going forward. Looking at the commit history, I can easily see the commits that shouldn't be there, as you say :).

PR #847, opened 2016-04-04
user name to alias

adhorrig:
Error message said "user name is required" when it was the alias field, this is easier to understand


Wardormeur:
key also used in https://github.com/CoderDojo/cp-zen-platform/blob/4507cfefb214f9d641e86db8c84bc65bac6af1a1/web/public/js/controllers/user-profile-controller.js#L283
I believe username message was allright on login controller, but isn't for alias (cf the file linked), so we should have 2 separate translations

adhorrig:
modified it, should make more sense now?

PR #849, opened 2016-04-05
775 : cannot invite certains usertYpes when owning multiple usertypes

Wardormeur:
Add start of permission service to restrict visibilities, actions, and data

Fix https://github.com/CoderDojo/community-platform/issues/775


adhorrig:
Tested, looks good!

PR #862, opened 2016-04-11
change getTotaLCount() to only get count for filtered users

butlerx:
change getTotaLCount() to only get count for filtered users or types instead of all users and types


Wardormeur:
Squash squash !

PR #865, opened 2016-04-15
Add Intercom debug info for specific bug

Wardormeur:


Wardormeur:
Not required anymore, but can serve for reference in case we want to link error popups with Intercom :)

PR #866, opened 2016-04-16
Fix for CoderDojo/community-platform#932

DanielBrierton:
Problem arose when you delete all the text from either input field on the time picker. The time picker returns null in this scenario, which when wrapped by moment results in "Invalid Date", which is from then on used as the basis for any time/date updates. The only way to kick out of this is to refresh the page.

I couldn't find any tests around this, but if there are some, point me in the right direction and I'll be happy to update them.

### DIFF
--- a/web/public/js/controllers/dojo-event-form-controller.js
+++ b/web/public/js/controllers/dojo-event-form-controller.js
@@ -79,6 +79,9 @@
   }
 
   function fixEventTime(newTime, date){
+    if (!newTime) {
+      return date;
+    }
     newTime = moment.utc(newTime);
     date = moment.utc(date);
     return moment.utc([ date.get('year'), date.get('month'), date.date(),


Wardormeur:
We have e2e tests (https://github.com/CoderDojo/cp-e2e-tests) but they're far from being up to date, so dw 
Will test today, thanks for this one, was looking the wrong way :+1:

DanielBrierton:
Cool, will check the e2e tests today. I've been wanting to try out Nightwatch for a while now, so this is as good an excuse as any.

Wardormeur:
Date value need to be also checked as when you modify post-null date, the old value of the watcher is still null

DanielBrierton:
I'm not sure I follow. A default value is set for fixedStartDateTime/fixedEndDateTime from the very start, and the only way those variables were becoming invalid was when newTime was null when calling fixEventTime. Now that fixEventTime just returns the existing fixedStartDateTime/fixedEndDateTime if newTime is null, fixedStartDateTime/fixedEndDateTime never get to an "Invalid Date" state.

If we were to check date in fixEventTime for future robustness' sake, what would we then return? I guess we could pass the defaultEventTime/defaultEventEndTime into the method when we call it and return that if both date and newTime are null.

Maybe I'm just misunderstanding your comment? Sorry if I am.

Wardormeur:
You're goddam right. I was trying with your changes deactivated, but since you return the new value as old value, it can only go wrong if the original date is wrong (which shouldn't happen)
As for a better return value, considering you're not able to submit the form without a value, which then is normally changed by the user manually, I think we don't really care :p

PR #873, opened 2016-04-26
User public view

butlerx:
add banner to say profile is public and add toggle to show what profile looks like publicly

### DIFF
--- a/web/public/templates/profiles/general-info.dust
+++ b/web/public/templates/profiles/general-info.dust
@@ -1,9 +1,11 @@
-<div ng-if="!edit-mode">
+<div ng-if="!editMode && !publicChampion">
   <div class="alert alert-info" ng-if="parentProfile && profile.ownProfileFlag">
-    {@i18n key="Only your username and profile picture are publicly viewable. Other fields are for administration purposes only."/}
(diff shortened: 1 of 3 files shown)


Wardormeur:
LGTM

web/public/templates/profiles/general-info.dust
```diff
@@ -1,3 +1,13 @@
+<div ng-if="!editMode && !publicChampion">
+  <div class="alert alert-info" ng-if="parentProfile && profile.ownProfileFlag">
+    {@i18n key="Only your username and profile picture are publicly viewable. Other fields are for administration purposes only. "/}
+    <a ng-click="publicProfile()">Click to view your public profile</a>
+  </div>
+  <div class="alert alert-info" ng-if="!parentProfile && (profile.ownProfileFlag  || profile.myChild)">
```
tangentfairy: This div is exactly the same as the one above - could the logic in both ng-ifs be combined so that the html is only written once?

Wardormeur: Probably, even more when the ng-if conditions are actually opposite :)
Will have a look at it this week, thanks for the head up !

PR #875, opened 2016-04-29
Export csv

butlerx:
Closes [#810](https://github.com/CoderDojo/community-platform/issues/810) 
and change how events exports works to use cdEventsService

### DIFF
--- a/lib/handlers.js
+++ b/lib/handlers.js
@@ -7,7 +7,7 @@ var debug = require('debug')('cp-zen-platform:handlers');
 
 module.exports = function(server, role) {
 
-  function doAct (request, reply, cmd, param, user) {
+  function doAct (request, reply, cmd, param, user, type) {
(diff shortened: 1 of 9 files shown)


lib/handlers.js
```diff
@@ -43,7 +43,9 @@ module.exports = function(server, role) {
         if (resp.http$.status) code = resp.http$.status;
         if (resp.http$.redirect) return reply.redirect(resp.http$.redirect);
       }
-      reply(resp).code(code);
+      if (type && type === 'csv') {
+        reply(resp.data).header('Content-Type', 'application/csv').header('Content-Disposition', 'attachment; filename=exported-list.csv');
```
Wardormeur: Can we make the filename more generic?

butlerx: what where you thinking?

PR #879, opened 2016-05-09
Change the flow for parent registration.

adhorrig:
Allows users to add to youths to their account from the initial registration form.
Also allows users to end up booking tickets after registering an account without having to go back to the Dojo Listing page. The flow is all built into one.

Original issue: https://github.com/CoderDojo/community-platform/issues/865


Wardormeur:
Will go through https://github.com/CoderDojo/cp-zen-platform/pull/887 , closing !

PR #909, opened 2016-06-07
Added right click support for dojo listings

lukekearney:
Hi,
I added the ability for people to right click dojo URLs  and open links in a new tab. I figured it would make things easier if people had multiple dojos in their area looking for volunteers

### DIFF
--- a/web/public/js/controllers/dojos-map-controller.js
+++ b/web/public/js/controllers/dojos-map-controller.js
@@ -2,9 +2,10 @@
 /* global google,jQuery,MarkerClusterer */
 
 //  TODO : reuse cd-dojos-map instead of this mixed-up controller
-function cdDojosMapCtrl($scope, $window, $state, $stateParams, $translate, $geolocation,  $location, cdDojoService, gmap, Geocoder, atomicNotifyService, usSpinnerService) {
+function cdDojosMapCtrl($scope, $window, $state, $stateParams, $translate, $geolocation,  $location, cdDojoService, gmap, Geocoder, atomicNotifyService, usSpinnerService, dojoUtils) {
   $scope.model = {};
   $scope.markers = [];
+  $scope.getDojoURL = dojoUtils.getDojoURL;
   var markerClusterer;
   var centerLocation = new google.maps.LatLng(25, -5);
 
@@ -33,7 +34,7 @@ function cdDojosMapCtrl($scope, $window, $state, $stateParams, $translate, $geol
 
   $scope.loadMap = function () {
     clearMarkers();
-    cdDojoService.list({verified: 1, deleted: 0, fields$:['name', 'geo_point', 'stage']}, function (dojos) {
+    cdDojoService.list({verified: 1, deleted: 0, fields$:['name', 'geo_point', 'stage', 'url_slug']}, function (dojos) {
       var filteredDojos = [];
       _.each(dojos, function (dojo) {
         if(dojo.stage !== 4) filteredDojos.push(dojo);
@@ -106,7 +107,7 @@ function cdDojosMapCtrl($scope, $window, $state, $stateParams, $translate, $geol
   $scope.getDojo = function (marker) {
     var dojoId = marker.dojoId;
     cdDojoService.load(dojoId, function (response) {
-      $scope.viewDojo(response);
+      $scope.getDojoURL(response);
     });
   };
 
@@ -230,4 +231,4 @@ function cdDojosMapCtrl($scope, $window, $state, $stateParams, $translate, $geol
 }
 
 angular.module('cpZenPlatform')
-  .controller('dojos-map-controller', ['$scope', '$window', '$state', '$stateParams', '$translate', '$geolocation', '$location', 'cdDojoService', 'gmap', 'Geocoder', 'atomicNotifyService', 'usSpinnerService', cdDojosMapCtrl]);
+  .controller('dojos-map-controller', ['$scope', '$window', '$state', '$stateParams', '$translate', '$geolocation', '$location', 'cdDojoService', 'gmap', 'Geocoder', 'atomicNotifyService', 'usSpinnerService', 'dojoUtils', cdDojosMapCtrl]);
--- a/web/public/js/controllers/user-profile-controller.js
+++ b/web/public/js/controllers/user-profile-controller.js
@@ -13,6 +13,9 @@ function cdUserProfileCtrl($scope, $rootScope, $state, $window, auth, cdUsersSer
 
   $scope.currentStateName = $state.current.name;
 
+  $scope.getDojoURL = dojoUtils.getDojoURL;
+
+
   if ($stateParams.showBannerMessage) {
     atomicNotifyService.info($translate.instant('Please complete your profile before continuing.'), 5000);
   }
--- a/web/public/js/directives/cd-dojos-map.js
+++ b/web/public/js/directives/cd-dojos-map.js
@@ -2,13 +2,13 @@
   'use strict';
   /*global google, StyledMarker, StyledIcon, StyledIconTypes */
 
-function cdDojosMap($stateParams, cdDojoService, $state) {
+function cdDojosMap($stateParams, cdDojoService, dojoUtils, $state) {
     return {
       restrict: 'E',
       template: '<div class="dojo-map row">'+
         '<section ng-if="mapLoaded" id="map">'+
         '<div ui-map-info-window="model.markerInfoWindow">'+
-          '<h3><a class="pointer" ng-click="getDojo(currentMarker)">{{ currentMarker.dojoName }}</a></h3>'+
+          '<h3><a class="pointer" ng-href="{{getDojoURL(currentMarker)}}">{{ currentMarker.dojoName }}</a></h3>'+
         '</div>'+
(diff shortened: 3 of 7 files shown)


DanielBrierton:
Hi @lukekearney,

Thanks for the pull request. This is something that was bugging me the other day, so it's extra appreciated! 😄 

Just a couple of issues:

1. The ng-href removes the need for the ng-click. In fact, Ctrl+Clicking or middle-clicking now opens a new tab, and the page in the current tab.
    * Also, class="pointer" isn't need now either, since it's a proper link.
2. You've set ng-href on an img tag, which doesn't work. We need to wrap the img in an a tag with the ng-href
3. For consistency, we should also change the behaviour of the popup for the map markers (it's in map-search.dust as well in the map section)

Thanks again!

lukekearney:
Hi @DanielBrierton 

I've removed the extra ng-click attributes and the ng-href from the image. The map should also work now with ng-href instead of ng-click

Wardormeur:
Allright, let's improve this even more :)
**Short way** :  don't repeat yourself and put getDojoURL in dojo-utils (a service) that you'll inject in every controller, and expose the function (instead of rewriting it)
**Longer way** : don't use ng-href, but ui-sref instead, and have viewDojo function (https://github.com/CoderDojo/cp-zen-platform/pull/909/files#diff-b3a6ae6e1d530cf7c5e2912f35e819d2L53) into the dojo without redirecting but returning the config object
This way you can use ui-sref as of ```ui-sref=dojo-detail(viewDojo(dojo))```
**Finally** : 
Verify that viewDojo is still used afterwards. If not, delete it !
And when you're done, squash it so we (try to) keep the branch clean (1 commit = 1 functionnality) (https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/ )

Thanks :+1:

lukekearney:
@Wardormeur I think that's it. I went with the injection method. viewDojo is still used by an image in user-profile.dust so I couldn't get rid of it, unfortunately

web/public/js/controllers/dojos-map-controller.js
```diff
@@ -103,10 +104,17 @@ function cdDojosMapCtrl($scope, $window, $state, $stateParams, $translate, $geol
     $state.go('dojo-detail',{country:country, path:path});
   }
 
+  $scope.getDojoURL = function(dojo) {
```
DanielBrierton: You've already set this function from dojoUtils, so you don't need this function definition anymore.

web/public/js/controllers/dojos-map-controller.js
```diff
@@ -103,10 +104,17 @@ function cdDojosMapCtrl($scope, $window, $state, $stateParams, $translate, $geol
...
+      return "/dojo/" + urlSlug;
+    }
+  }
+
   $scope.getDojo = function (marker) {
     var dojoId = marker.dojoId;
-    cdDojoService.load(dojoId, function (response) {
-      $scope.viewDojo(response);
+    return cdDojoService.load(dojoId, function (response) {
+      return $scope.getDojoURL(response);
```
DanielBrierton: Unfortunately this won't work as the getDojo function will have already returned the response of cdDojoService.load before the Dojo details have even loaded.

I think we may actually need to leave the map links as they are for now, as we need to asynchronously get the Dojo URL at the time the marker is selected.

web/public/js/controllers/user-profile-controller.js
```diff
@@ -13,6 +13,9 @@ function cdUserProfileCtrl($scope, $rootScope, $state, $window, auth, cdUsersSer
 
   $scope.currentStateName = $state.current.name;
 
+  $scope.getDojoURL = dojoUtils.getDojoURL;
```
DanielBrierton: You're using dojoUtils but haven't injected it here.

web/public/js/controllers/user-profile-controller.js
```diff
@@ -491,6 +494,13 @@ function cdUserProfileCtrl($scope, $rootScope, $state, $window, auth, cdUsersSer
     $state.go('dojo-detail',{country:country, path:path});
   }
 
+  $scope.getDojoURL = function(dojo) {
```
DanielBrierton: Again, you've already set this. There's no need to write it again.

web/public/js/directives/cd-dojos-map.js
```diff
@@ -58,6 +59,14 @@ function cdDojosMap($stateParams, cdDojoService, $state) {
             var path = urlSlugArray.join('/');
             $state.go('dojo-detail', {country:country, path:path });
           }
+
+          $scope.getDojoURL = function(dojo) {
```
DanielBrierton: You should inject dojoUtils in this file too and use it here instead of writing a function just for this file.

web/public/js/services/dojo-utils.js
```diff
@@ -50,5 +50,13 @@ angular.module('cpZenPlatform').factory('dojoUtils', function($location, $transl
       });
     }
   };
+
+  dojoUtils.getDojoURL = function(dojo) {
+    if (dojo) {
+      console.log(dojo);
```
DanielBrierton: Remove the console.log

web/public/templates/dojos/map-search.dust
```diff
@@ -1,7 +1,7 @@
 
   <section ng-if="mapLoaded" id="map">
     <div ui-map-info-window="model.markerInfoWindow">
-      <h3><a class="pointer" ng-click="getDojo(currentMarker)">{{ currentMarker.dojoName }}</a></h3>
+      <h3><a ng-href="{{ getDojoURL(currentMarker.dojo) }}">{{ currentMarker.dojoName }}</a></h3>
```
DanielBrierton: Probably revert this for now, along with the changes to the getDojo method.

web/public/templates/dojos/user-profile.dust
```diff
@@ -129,7 +129,7 @@
             <div class="dojo-container" ng-repeat="dojo in profile.dojos">
               <img ng-click="viewDojo(dojo)" src="/img/dojo-default-logo.png" class="dojo-thumbnail pointer">
```
DanielBrierton: You can wrap this img in an <a> tag with ng-href instead of using ng-click on the img. This will then provide a consistent behaviour for both the image and link.

lukekearney:
@DanielBrierton I think I just rebased incorrectly or otherwise lost some of the changes I made with the service injection etc. Fixed now.

The map appears to work fine with ng-href. It opens a tooltip with a link to the dojo page

DanielBrierton:
Ah, cool. I wasn't actually sure if getDojoUrl could work off the marker data, but that's cool!

My comments on user-profile-controller.js and cd-dojos-map.js still need to be addressed though.
* https://github.com/CoderDojo/cp-zen-platform/pull/909/files#r66220177
* https://github.com/CoderDojo/cp-zen-platform/pull/909/files#r66220201

Other than that, it looks good. Once you've made those changes, we just need to sort out your branch so we can have a nice clean merge to master. What I recommend doing at this point is to fetch upstream, then reset --soft to upstream/master
```
git fetch upstream
git reset --soft upstream/master
```
(You may need to change upstream to whatever you called the remote for upstream)
This will stage your changes on top of a clean master. You can then just commit as normal, and you'll end up with just one commit with all your changes.

If you have any trouble with that, you can pop over onto our Gitter and I can talk you through it. https://gitter.im/CoderDojo/community-platform

You're nearly there! 👍  Thanks again!

lukekearney:
That should be everything as outlined above. Let me know if there are any issues

DanielBrierton:
I've been testing this, and just found an issue in our embeddable Dojo map. There's a template inside cd-dojos-map.js that needs to be updated to use getDojoURL as well. As well as that, there's a typo in cd-dojos-map's dependencies. You have dojoUrils, plus the order of dependencies in the array doesn't match the order in the callback.

You can test this directive at http://localhost:8000/embedded/dojos-map/lat/0/lon/0?zoom when running the stack

Last thing then is just that you still have a console.log in the getDojoURL method itself.

lukekearney:
The embedded map should be consistent now. Didn't even notice the typo when I was testing it on my own machine so thanks for finding it, but that's fixed as well

DanielBrierton:
Yeah, the typo only affected the embeddable map, which is probably why you didn't see it. cd-dojos-map isn't the same map as the home page (although it probably should be, that's a task for another day).

I've read over the code and tested it, and you'll be happy know I haven't found any further issues.

@Wardormeur, could you take a quick look over this before we merge it in?

web/public/js/directives/cd-dojos-map.js
```diff
@@ -35,6 +36,7 @@ function cdDojosMap($stateParams, cdDojoService, $state) {
                     map: $scope.model.map,
                     dojoName: dojo.name,
                     dojoId: dojo.id,
+                    urlSlug: dojo.urlSlug,
```
Wardormeur: Waaaaay better, thanks :+1:

Wardormeur:
LGTM, congratz ! :)

DanielBrierton:
Sorry for the delay in this comment, but thanks again for the PR! It's now deployed to production and working perfectly!

If you ever feel like contributing more and have any questions, don't hesitate to ask, we're always happy to help.

PR #913, opened 2016-06-16
Moved email subject translation to back-end

DanielBrierton:


DanielBrierton:
Requires CoderDojo/cp-dojos-service#265 and CoderDojo/cp-events-service#913

Wardormeur:
See https://github.com/CoderDojo/cp-dojos-service/pull/265/files#r67357425, i believe moving the subject to the backend is necessary

web/index.js
```diff
@@ -295,18 +295,12 @@ function formatLocaleCode (code) {
 }
 
 var locality = function(request) {
-  var ngKey;
-  if (request.state.NG_TRANSLATE_LANG_KEY) {
-    // TODO - shouldn't hapi decode cookies?
-    ngKey = decodeURIComponent(request.state.NG_TRANSLATE_LANG_KEY);
-    ngKey = ngKey.replace(/\"/g, '');
-  }
+  var localesFormReq = (request.state && request.state.NG_TRANSLATE_LANG_KEY && request.state.NG_TRANSLATE_LANG_KEY.replace(/\"/g, ''))
```
Wardormeur: Not sure to get what happend here :)

web/public/js/controllers/accept-session-invite-controller.js
```diff
@@ -13,9 +13,8 @@
       ticketId: ticketId,
       invitedUserId: invitedUserId,
       emailSubject: {
-        'request':  $translate.instant('Your ticket request for'),
```
Wardormeur: missing on purpose?

web/public/js/controllers/manage-event-applications-controller.js
```diff
@@ -385,10 +385,9 @@
 
       application = _.omit(application, ['user', 'age', 'parents', 'dateApplied', 'applicationDates', 'attendanceModel']);
       application.emailSubject = {
-        'request':  $translate.instant('Your ticket request for'),
```
Wardormeur: deleted on purpose?

web/public/js/controllers/accept-session-invite-controller.js
DanielBrierton: Yup. No need for it anymore. This piece of the subject has been rolled into other properties.

web/public/js/controllers/manage-event-applications-controller.js
DanielBrierton: Same as the other one, yeah, this is on purpose.

DanielBrierton:
Re: your comment on moving subject to backend

I opted to take a short route here, and just stop translating on client-side and just send the key down. If the key passed down doesn't exist, the email won't be sent. We could absolutely remove this entirely from here, and do it lower in the stack. Shouldn't be a big thing, just need to figure out the exact flows through the code for emails.

Wardormeur:
Well, it's more likely thatI think a key doesn"t exists in the specified language cause it to says the subject is invalid. And i believe a non-translated key while then make sending of email impossible as they don't exists (i believe) until those are translated

DanielBrierton:
No, it'll default to en_US if it doesn't find the key in the specified locale. If the key doesn't exist in en_US, then the email will fail.

Wardormeur:
didn't notice https://github.com/CoderDojo/cp-dojos-service/pull/265/files#diff-acc92eaa186b23298acbfa3e913a7202R41, my bad :-1:

Wardormeur:
Security-wise it doens't remvoe the fact that you can specify another key (eg reset password) while having a different body
Is that Okaysh ?

Wardormeur:
LGTM

PR #915, opened 2016-06-20
Prevent user from scaling page since we're responsive

DanielBrierton:
Fix for CoderDojo/community-platform#978


Wardormeur:
Bootstrap recommend using ```<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no">``` with not scalable, could be something to consider?
http://getbootstrap.com/css/#overview-mobile

DanielBrierton:
It shouldn't really make any difference, but I've pushed the change anyway.

PR #916, opened 2016-06-20
Bugfix/securitypatchfix issue

Wardormeur:


lib/dojos.js
```diff
@@ -1078,25 +1078,6 @@ exports.register = function (server, options, next) {
     }
   }, {
     method: 'GET',
-    path: options.basePath + '/countries/continents/lat-long',
```
DanielBrierton: Purposely removed I assume?

lib/dojos.js
```diff
@@ -1134,25 +1115,6 @@ exports.register = function (server, options, next) {
       }
     }
   }, {
-    method: 'GET',
```
DanielBrierton: And again, purposely removed?

lib/dojos.js
Wardormeur: unused y

lib/dojos.js
Wardormeur: unused y

PR #918, opened 2016-06-22
Fix API point to what it was supposed to be

Wardormeur:
Goes with https://github.com/CoderDojo/cp-badges-service/pull/55


DanielBrierton:
LGTM

PR #919, opened 2016-06-28
remove unused and unneeded functuins and use user for mentor check

butlerx:


DanielBrierton:
LGTM

Wardormeur:
We should refacto cp-perms to allow us to check upon an userType instead of having it "in" the test workflow
LGTM anyway

PR #921, opened 2016-06-29
make map open at current location

butlerx:
closes [#948](https://github.com/CoderDojo/community-platform/issues/948)

### DIFF
--- a/web/public/js/controllers/dojos-map-controller.js
+++ b/web/public/js/controllers/dojos-map-controller.js
@@ -2,12 +2,13 @@
 /* global google,jQuery,MarkerClusterer */
 
 //  TODO : reuse cd-dojos-map instead of this mixed-up controller
-function cdDojosMapCtrl($scope, $window, $state, $stateParams, $translate, $geolocation,  $location, cdDojoService, gmap, Geocoder, atomicNotifyService, usSpinnerService, dojoUtils) {
+function cdDojosMapCtrl($scope, $window, $state, $stateParams, $translate, $geolocation, $q, $location, cdDojoService, gmap, Geocoder, atomicNotifyService, usSpinnerService, dojoUtils) {
   $scope.model = {};
   $scope.markers = [];
   $scope.getDojoURL = dojoUtils.getDojoURL;
   var markerClusterer;
   var centerLocation = new google.maps.LatLng(25, -5);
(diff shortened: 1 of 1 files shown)


web/public/js/controllers/dojos-map-controller.js
```diff
@@ -43,11 +43,27 @@ function cdDojosMapCtrl($scope, $window, $state, $stateParams, $translate, $geol
...
     }
+
+    if (navigator.geolocation) {
+      navigator.geolocation.getCurrentPosition(function(position) {
+        var pos = {
+          lat: position.coords.latitude,
+          lng: position.coords.longitude
+        };
+        $scope.model.map.setCenter(pos);
+        $scope.model.map.setZoom(6);
```
DanielBrierton: Should we perhaps increase the zoom level some more? I like 11. It seems to be close enough to get a good view of the surrounding area, and not too much pointless data further out.

Wardormeur: Reason for 6 was because if there is no dojo in your area, it looks like a crapilly empty map @DanielBrierton

butlerx: i can make it 7 or 8?

web/public/js/controllers/dojos-map-controller.js
```diff
@@ -198,22 +222,40 @@ function cdDojosMapCtrl($scope, $window, $state, $stateParams, $translate, $geol
...
+            dojoName: dojo.name,
+            dojoId: dojo.id,
+            icon: 'img/marker' + pinColor +'.png',
+            position: new google.maps.LatLng(dojo.geoPoint && dojo.geoPoint.lat || dojo.geo_point.lat, dojo.geoPoint && dojo.geoPoint.lon || dojo.geo_point.lon)
+          });
+          marker.dojo = dojo;
+          $scope.markers.push(marker);
+        }
+      });
+      markerClusterer = new MarkerClusterer($scope.model.map, $scope.markers, {imagePath:'components/google-maps-utility-library-v3/markerclusterer/images/m'});
```
Wardormeur: if you return, this will never be executed?

DanielBrierton: It will. It returns the Promise, which when chained with other promises will then get executed.

PR #925, opened 2016-07-01
Allow kids to use projects in their profiles

Wardormeur:


butlerx:
LGTM

Wardormeur:
May close actually, a whole rework of projects may be needed (better structure(name/link/language)) to go with the badge linking

DanielBrierton:
Yeah, might be a good idea not to do this. Otherwise we're gonna have a free text field that we're not going to be able to migrate to our new projects structure when that comes. Unless we were to show the old the free text field until the user adds projects through the new functionality.

PR #929, opened 2016-07-08
LMS Integration

Wardormeur:


lib/users.js
```diff
@@ -245,6 +245,36 @@ exports.register = function (server, options, next) {
...
+          responseMessages: [
+            {code: 200, message: 'OK'}
+          ]
+        }
+      }
+    }
+  }, {
+    method: 'POST',
+    path: options.basePath + '/lms/badge',
+    handler: actHandlerCheckCertificate,
```
DanielBrierton: Not really descriptive of what it actually does. Perhaps should be specific, or take an argument for the act executed after checking the cert.

Wardormeur: Not sure to get what you mean, it's returning the link used for the front-end

Wardormeur: explained irl, got it :)

DanielBrierton:
Tested, and all works well. The only issue I found was that because I didn't add any tags to my badge for it, it didn't show on my profile. Perhaps we should have an Other section for those that don't match the tags we look for?

PR #932, opened 2016-07-11
Initial implementation of GA

Wardormeur:
send event and navigation to GA by extending ngClick behavior. Requires ngAria-label to be used, see example of implementation in /templates
TODO : use gulp to verify such attribute exists


web/lib/dust-loadjs.js
```diff
@@ -20,8 +20,19 @@ if (process.env.UIDEBUG === 'true') {
...
+  } else {
+    conf = require('../config/web-development.js');
+  }
+  for (var variable in conf) {
+    js += 'window.zenConf.' + variable + ' = \'' + conf[variable] +'\';';
+  }
   for (var i = 0; i < scripts.length; i++) {
     scriptTags += '<script src="' + scripts[i] + '"></script>';
   }
+  scriptTags += '<script>' + js + '</script>';
```
DanielBrierton: Should we not put this above the other scripts so it's there when they run?

PR #940, opened 2016-08-11
Updated so profile menu works when running locally

DanielBrierton:


DanielBrierton:
Need to re-run bower install to get latest cd-menu for fix.

PR #947, opened 2016-08-18
Added picture grid directive

DanielBrierton:


butlerx:
LGTM

web/public/js/directives/cd-picture-grid/template.dust
```diff
@@ -0,0 +1,6 @@
+<div class="cd-picture-grid">
+    <a class="cd-picture-grid__item" ng-repeat="item in $ctrl.items" href="{{ item.href }}">
```
Wardormeur: should'nt it be ng-href?

web/public/js/directives/cd-picture-grid/template.dust
```diff
@@ -0,0 +1,6 @@
...
+        <div class="cd-picture-grid__picture" style="background-image: url({{ item.picture }});"></div>
```
Wardormeur: same here, ng-style?

web/public/js/directives/cd-picture-grid/template.dust
DanielBrierton: Fair enough, changed that now, will push in a min

web/public/js/directives/cd-picture-grid/template.dust
DanielBrierton: ng-style is for applying styles from objects in the scope. The items array would then have to be:
```
[{
  href: '/whatever',
  caption: 'Whatever',
  picture: {
    'background-image': 'url(/whatever.png)'
  }
}]
```

Seems easier for users of the directive to leave it as is.

Wardormeur: Fair, forgot ng-style was oriented on optional styling (the same as ng-class)

PR #949, opened 2016-08-18
Fixed issue with adding content dynamically to card

DanielBrierton:


web/public/js/directives/cd-expanding-card/index.js
```diff
@@ -35,14 +35,18 @@
           //content.css('height', 'auto');
           content.css('height', contentWrapper.outerHeight() + 'px');
           $timeout(function () {
+            content.css('height', 'auto');
             ctrl.expanded = 'expanded';
           }, 300);
         } else if (force) {
```
Wardormeur: means you can't have both "collapsed" and "force"? 
In which case, 2 fn would be nicer?

DanielBrierton: I'm not following what you mean...

Wardormeur:
LGTM :+1:

PR #960, opened 2016-08-23
Restrict visibility of email

Wardormeur:
Frankly backend doesn't send the email back, so it's just a preliminary fix of "how it should work"

### DIFF
--- a/web/public/js/controllers/user-profile-controller.js
+++ b/web/public/js/controllers/user-profile-controller.js
@@ -27,6 +27,7 @@ function cdUserProfileCtrl($scope, $rootScope, $state, $window, auth, cdUsersSer
   $scope.editMode = false;
   $scope.publicMode = false;
   $scope.publicChampion = false;
+  $scope.loggedInUserIsMemberOfDojoChampion = false;
   var profileUserId = $state.params.userId;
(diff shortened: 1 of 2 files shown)


web/public/js/controllers/user-profile-controller.js
```diff
@@ -27,6 +27,7 @@ function cdUserProfileCtrl($scope, $rootScope, $state, $window, auth, cdUsersSer
   $scope.editMode = false;
   $scope.publicMode = false;
   $scope.publicChampion = false;
+  $scope.loggedInUserIsMemberOfDojoChampion = false;
```
DanielBrierton: loggedInUserIsDojoChampion?

Wardormeur: Nope, you want to check the opposite, if the loggedInUser is a member of the profile's dojo he's viewing

DanielBrierton: To be exact, is it checking that the user's profile being viewed is a member of a dojo that the currently logged in user is a champion of?

Wardormeur: Nope, if the profile being viewed is a champion for the dojo that the user seeing is belonging to

PR #961, opened 2016-08-23
Ui refresh loading icon when reloading

Wardormeur:


butlerx:
not a big fan of the var name but LGTM

PR #962, opened 2016-08-23
Ui refresh fix profile stretching

Wardormeur:


web/public/js/directives/cd-rounded-image/style.less
```diff
@@ -5,6 +5,10 @@
     border: none;
     cursor: pointer;
     z-index: 2;
+    object-fit: cover;
```
butlerx: cool feature

butlerx:
this seems to have parts of #961

Wardormeur:
They all have, but github does the job once the parent commit is pushed ;)

PR #964, opened 2016-08-24
Fix SAD process

Wardormeur:
Still ugly tho
hopefully doesn't break anything else

### DIFF
--- a/web/public/css/header.less
+++ b/web/public/css/header.less
@@ -133,26 +133,26 @@
 }
 
 .cd-color-1, html .container .cd-color-1 {
-  background-color: #61C93F;
+  background-color: #61C93F !important;
 }
 
 .cd-color-2, html .container .cd-color-2 {
-  background-color: #2E7AC7;
+  background-color: #2E7AC7 !important;
 }
 
 .cd-color-3, html .container .cd-color-3 {
-  background-color: #FBCC33;
+  background-color: #FBCC33 !important;
 }
 
 .cd-color-4, html .container .cd-color-4 {
-  background-color: #ED462F;
+  background-color: #ED462F !important;
 }
 
 .cd-color-5, html .container .cd-color-5 {
-  background-color: #2C9CFB;
+  background-color: #2C9CFB !important;
 }
 .cd-color-6, html .container .cd-color-6 {
-  background-color: #FFA500;
+  background-color: #FFA500 !important;
 }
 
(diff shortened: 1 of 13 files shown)


web/public/css/header.less
```diff
@@ -133,26 +133,26 @@
 }
 
 .cd-color-1, html .container .cd-color-1 {
-  background-color: #61C93F;
+  background-color: #61C93F !important;
```
butlerx: why :question:

web/public/js/controllers/start-dojo-wizard-controller.js
```diff
@@ -232,8 +240,10 @@ function startDojoWizardCtrl($scope, $window, $state, $location, auth, alertServ
   }
 
   $scope.disabled = true;
-  $scope.next = function () {
+  $scope.next = function (form) {
     $scope.disabled = false;
+    //  This is a pure hack to escape the scope issue, i'm sorry for the next guy. (G.F.)
```
butlerx: you'll probably end up being the next guy

butlerx:
Tested works for logged in users and new users and doesnt break separate reg flow for user only

web/public/css/header.less
DanielBrierton: ❓ ❔ ❗ ❕ ❓ ❔ ❗ ❕

Wardormeur: Because it's used for panels in the "setup your dojo" step of the SAD process and it's overwritten by bootstrap which has a more strict rule ( panel>heading)
Best solution would be to do a loop for colors inside a mixin, but since SAD process must be reworked graphically anyway, that's a quickfix

butlerx: so hacky hack is hack?

Wardormeur: That's more like "I make it work, the redesign of this process is supposed to be in this sprint and we're 2 days late from the initial deployment" :)

DanielBrierton: Fine, I suppose. I better see `!important` in red in the SAD pull request. 😛

Wardormeur: The whole cd-color-n won't be there anymore dw ;)

web/public/js/controllers/start-dojo-wizard-controller.js
DanielBrierton: 🙏

PR #965, opened 2016-08-24
Add loading spinner when refreshing

Wardormeur:


Wardormeur:
fucked up git from #961

PR #971, opened 2016-08-25
Ui Fixes

Wardormeur:
Centering of Claim badge button
Display children list only if adult (was commented out for debug purpose, leftover..)


DanielBrierton:
👍

PR #975, opened 2016-08-25
Added page for getting currently logged in profile

DanielBrierton:
Needed for CoderDojo/nodebb-theme-cd-persona#4

### DIFF
--- a/web/public/js/init-master.js
+++ b/web/public/js/init-master.js
@@ -105,8 +105,26 @@
     championsForUser: function ($stateParams, cdUsersService) {
       return cdUsersService.loadChampionsForUserPromise($stateParams.userId).then(winCb, failCb);
     },
+    championsForLoggedInUser: function ($stateParams, cdUsersService, auth) {
+      return auth.get_loggedin_user_promise().then(function (currentUser) {
+        if(currentUser){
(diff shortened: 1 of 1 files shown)


butlerx:
:+1:

web/public/js/init-master.js
```diff
@@ -102,11 +102,31 @@
     hiddenFields: function(cdUsersService){
       return cdUsersService.getHiddenFieldsPromise().then(winCb, failCb);
     },
-    championsForUser: function ($stateParams, cdUsersService) {
-      return cdUsersService.loadChampionsForUserPromise($stateParams.userId).then(winCb, failCb);
+    championsForUser: function ($stateParams, cdUsersService, auth) {
```
Wardormeur: I 'd prefer to decuplate the resolve as they don't serve the same data :  it was "get the champions of the visible profile" and now it's also "get the champion of the connected user"?

web/public/js/init-master.js
```diff
@@ -643,6 +673,26 @@
           },
           templateUrl: '/directives/tpl/user/cd-profile/edit'
         })
+        .state("my-profile", {
+          url: '/profile',
```
Wardormeur: missing "public" param :)

PR #981, opened 2016-08-30
Added action bar directive

DanielBrierton:


web/public/js/directives/cd-action-bar/style.less
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }
+
+  position: fixed;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  background: @cd-blue;
+  height: 40px;
+  z-index: 5000;
```
Wardormeur: We should probably start a list of z-index to avoid overlap issues..

DanielBrierton: Yes, I concur.

Wardormeur:
@DanielBrierton close please :D

PR #985, opened 2016-08-31
small front-end fixes

Wardormeur:
Change date format for recurrent event
change ribbon to not overlap the banner anymore && added box-shadow
start harmonizing colors by importing cd-commons > must be discussed, tint seems to make quite.. darkish colors
remove border-radius for share-event

### DIFF
--- a/web/public/js/directives/event/list/item/index.js
+++ b/web/public/js/directives/event/list/item/index.js
@@ -16,6 +16,7 @@ angular
             cdELI.isEmbedded = embedder.isEmbedded();
             cdELI.event.isPast = eventUtils.isEventInPast(_.last(cdELI.event.dates));
             cdELI.event = eventUtils.getFormattedDates(cdELI.event);
+            cdELI.datesExpanded = false;
             cdELI.goTo = function() {
               embedder.redirectWrapper( function(){
                 $state.go('dojo-event-details', {dojoId: cdELI.event.dojoId, eventId: cdELI.event.id});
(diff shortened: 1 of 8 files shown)


web/public/js/directives/event/list/item/index.js
```diff
@@ -16,6 +16,22 @@ angular
             cdELI.isEmbedded = embedder.isEmbedded();
             cdELI.event.isPast = eventUtils.isEventInPast(_.last(cdELI.event.dates));
             cdELI.event = eventUtils.getFormattedDates(cdELI.event);
+            if (!cdELI.event.isPast && cdELI.event.type === 'recurring') {
```
DanielBrierton: Couldn't this whole block be done just by:

```
cdELI.event.upcomingDates = cdLI.event.dates.filter(function (date) {
  return !eventUtils.isEventInPast(date);
});
cdELI.event.nextDate = moment.utc(_.first(cdELI.event.upcomingDates).startTime).format('Do MMMM YY');
```

butlerx:
:100: / :100:

web/public/js/directives/event/list/item/index.js
Wardormeur: we still need the formattedDates which are != from the dates
Plus the while loop is there to ensure to stop when we find the first one. The reason is : over a year of repeated event, it can make a difference perfomance-wise than doing it over every date.

DanielBrierton:
Other than my comment [here](https://github.com/CoderDojo/cp-zen-platform/pull/985#discussion_r77517675), I'm happy. 👍

web/public/js/directives/event/list/item/index.js
DanielBrierton: Ok, makes sense. 👍  from me so!

PR #986, opened 2016-08-31
Fix icon margin in button for button with more than icon

Wardormeur:


butlerx:
:+1:

web/public/css/base/_button.less
```diff
@@ -10,7 +10,7 @@
 }
 
 .btn {
-  .fa {
+  .fa:not(:only-child){
```
DanielBrierton: TIL 👍

PR #988, opened 2016-09-01
Remove checksum verification until they fixes it

Wardormeur:


DanielBrierton:
(sad panda)

Please don't abuse kthx.

PR #989, opened 2016-09-05
Picture grid actions

DanielBrierton:


web/public/js/directives/cd-action-bar/cd-action-bar-item/style.less
```diff
@@ -0,0 +1,17 @@
...
+    transition: background 0.2s ease-out;
+
+    .fa {
+      margin: 0 6px;
+    }
+
+    &:hover, &:active {
+      color: @cd-white;
+      text-decoration: none;
+      background: tint(@cd-blue, 10%);
```
Wardormeur: can use cd-common

web/public/js/directives/cd-action-bar/style.less
```diff
@@ -0,0 +1,88 @@
...
+
+    to {
+      transform: translateY(100%);
+    }
+  }
+
+  background: @cd-blue;
+  height: 40px;
+  z-index: 5000;
+  max-width: 1280px;
```
Wardormeur: max-width : @screen-md-max?

web/public/js/directives/cd-action-bar/style.less
```diff
@@ -0,0 +1,88 @@
...
+  margin: 0 auto;
+  display: flex;
+  transform: translateY(100%);
+  animation-fill-mode: both;
+  padding-bottom: 40px;
+  margin-bottom: -40px;
```
Wardormeur: same hax than mine before, no issues so far? (un-clickable zone on responsive?)

web/public/js/directives/cd-action-bar/cd-action-bar-item/style.less
DanielBrierton: Need #985 merged before being able to do this, but yeah, we should for sure.

web/public/js/directives/cd-picture-grid/index.js
```diff
@@ -3,8 +3,37 @@
...
+      ctrl.selectedItems = [];
+      var gridItems = null;
+      var selectedClass = 'cd-picture-grid__item--selected';
+
+      ctrl.handleSelection = function (e, item) {
+        if (ctrl.actions) {
+          e.preventDefault();
+          var $el = $(e.currentTarget);
+          if (ctrl.multipleSelection !== true && ctrl.selectedItems.length > 0) {
+            gridItems = gridItems || $element.find('.cd-picture-grid__item'); // so we only query the DOM once
```
Wardormeur: pagination won't be an issue w/ that?

web/public/js/directives/cd-action-bar/style.less
DanielBrierton: Similar hack, different reason. This is just so when the action bar bounces up that there's no white gap underneath it. For this, it's either fixed to the bottom, so a minus bottom margin going off the page, or is underneath the footer thanks to z-index and therefore doesn't affect click-ability.

Wardormeur: Great, thanks :)

web/public/js/directives/cd-picture-grid/index.js
DanielBrierton: Should probably build pagination into this directive to make it easier to deal with pagination. At the moment, yeah, there's no real support for pagination with multi-selection here...

butlerx: might infinite scroll be an idea when you have multiple users selected so you can keep scrolling?

Wardormeur: eh, whatever, i was just thinking that the lack of refresh was blocking it, but since we're paginated everywhere, we'll consider it's only page/page.
Will it refresh the "selection" post-new page?

DanielBrierton: It does now! 😛

PR #992, opened 2016-09-05
Fix export csv

Wardormeur:


butlerx:
:rocket: :tada: :+1:

web/public/js/controllers/manage-event-applications-controller.js
```diff
@@ -86,11 +86,22 @@
...
         var downloadLink = angular.element('<a></a>');
-        var csv = new Blob([response], { type: "text/csv;charset=utf-8;" });
-
-        downloadLink.attr('href',(window.URL || window.webkitURL).createObjectURL(csv));
-        window.open(downloadLink[0]);
+        var csv = new Blob([response.data], { type: "text/csv;charset=utf-8;" });
+        var urlFactory = window.URL || window.webkitURL;
+        var tempFile = urlFactory.createObjectURL(csv);
+        downloadLink.attr('href', tempFile);
+        downloadLink.attr('download', $scope.event.name + '_' + $translate.instant('Applicants') + moment().format() + '.csv');
```
DanielBrierton: A little concerned that a translation of Applicants could have invalid filename characters. I'd imagine browsers will handle this gracefully though...

Wardormeur: Actually, the dojo name is even more dangerous ;)
Tho, it seems createObjectURL does URI encode by default, so that it's still usable. We could encoreURI the translation just to be sure ?

DanielBrierton: Just did a quick test, and the browser replaces invalid characters with dashes, so 👍 , no need to URI encode

PR #994, opened 2016-09-05
ui bugfix : recurring event for detail page, default space for buttons w/ icons

Wardormeur:
Refactored code from last week to apply the same behavior to event detail (TODO : real UI, it's ugly AF, remove the popup plz)
Remove original catchall for .fa inside buttons to replace it with a hack that shouldn't impact anybody but leave it with a default space when the class is missing. TODO : add the real margin by adding the class to the necessary buttons

### DIFF
--- a/web/public/js/services/event-utils.js
+++ b/web/public/js/services/event-utils.js
@@ -25,7 +25,7 @@ angular.module('cpZenPlatform').factory('eventUtils', ['$translate', function($t
       return formattedDates.slice(nextDateIndex);
     }
     // Congrats, you called this function upon an past event
-    return false;
+    return [];
   }
 
   eventUtils.getFormattedDates = function (event) {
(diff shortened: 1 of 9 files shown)


web/public/js/services/event-utils.js
```diff
@@ -2,15 +2,33 @@
...
+        nextDateIndex = eventIndex;
+      }
+      eventIndex +=1;
+    }
+
+    if (nextDateIndex >= 0 && formattedDates) {
+      return formattedDates.slice(nextDateIndex);
+    }
+    // Congrats, you called this function upon an past event
+    return false;
```
DanielBrierton: I'd rather `return [];` here since we otherwise return an array of dates. Makes more sense to go `_.isEmpty(nextDates)` than `nextDates === false` to me.

butlerx: id agree

DanielBrierton:
👍

butlerx:
:fire: away

PR #996, opened 2016-09-06
Added Verified At column to manage dojos page

DanielBrierton:


butlerx:
:100:% :fire: away. good to :rocket:

PR #1003, opened 2016-09-13
Fix order of events, display of next date for recurring events

DanielBrierton:


web/public/js/controllers/dojo-events-list-controller.js
```diff
@@ -70,6 +70,19 @@
           event.for = $translate.instant(userType);
           events.push(event);
         });
+        events.sort(function (eventA, eventB) {
```
Wardormeur: tempted to say we should move it to event utils, as it's probably going to be used for my-tickets ?

Wardormeur:
LGTM

PR #1007, opened 2016-09-16
Added i18n fallbacks

DanielBrierton:


web/public/js/init-master.js
```diff
@@ -859,6 +859,18 @@
...
      function($rootScope, $filter, $state, embedder, $cookieStore, $document, verifyProfileComplete, alertService, $translate, $location){
 
+      // Override $translate.instant so it falls back to en_US, then the original key when no result
+      var originalTranslateInstant = $translate.instant;
+      $translate.instant = function (key) {
+        var translation = originalTranslateInstant.apply($translate, arguments);
+        if (!translation) {
+          var args = Array.prototype.slice.apply(arguments); // Needed so we can modify the arguments
+          args[3] = 'en_US'; // The forceLanguage argument
+          translation = originalTranslateInstant.apply($translate, args);
```
DanielBrierton: I've actually found a problem where this is just returning an empty string even if the translation exists in en_US... I'm investigating now.

DanielBrierton: Ok, that was because I forgot to commit the file that returns the dictionaries based on the URI param.

On your comment then, yeah, if the key doesn't exist, then translation will become the key here, but if the key does exist, but for some reason the value is an empty string in the en_US messages.po, then we want to default to the key, hence the `|| key`

web/lib/dust-i18n.js
```diff
@@ -31,13 +21,18 @@ dust.helpers.i18n = function (chunk, context, bodies, params) {
...
+  var translation = i18nHelper.getClosestTranslation(locale, params.key);
+  if (translation) {
+    if (params.context) {
+      translation = translation.withContext(params.context);
+    }
+    translation = params.count
+      ? translation.ifPlural(params.count, params.key).fetch(params.count)
+      : translation.fetch();
+  } else {
+    // console.log('Missing translation', params.key); // Uncomment this to track down missing messages.po entries used in templates
```
DanielBrierton: Not sure I'd want to log this anyway, as I fear it would be too much noise in our logs. The intention here was for a manual way to track down missing keys. We don't have access to `server` anyway

PR #1014, opened 2016-10-10
Limit selection of users to avoid overbooking

Wardormeur:
Ui-wise, it's not perfect, but it should do the trick while we wait for our renewal of this "popup"

### DIFF
--- a/web/locale/en_US/messages.po
+++ b/web/locale/en_US/messages.po
@@ -2479,6 +2479,11 @@ msgstr "Myself"
 msgid "Apply for"
 msgstr "Apply for"
 
+msgid "Tickets left"
+msgstr "Tickets left"
+
+msgid "Apply for"
+msgstr "Apply for"
 msgid "Export to CSV"
 msgstr "Export to CSV"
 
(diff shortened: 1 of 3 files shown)


web/locale/en_US/messages.po
```diff
@@ -2479,6 +2479,11 @@ msgstr "Myself"
 msgid "Apply for"
 msgstr "Apply for"
 
+msgid "Tickets left"
+msgstr "Tickets left"
+
+msgid "Apply for"
```
DanielBrierton: Duplicate string

web/public/js/controllers/session-modal-controller.js
```diff
@@ -38,6 +40,9 @@
     _.each(_.keys($scope.session.tickets), function(ticketId) {
       var ticket = $scope.session.tickets[ticketId];
       $scope.session.tickets[ticketId].remaining = ticket.quantity - ticket.totalApplications;
+      var settings = _.clone($scope.applyForSettings);
+      settings.selectionLimit = ticket.remaining;
```
DanielBrierton: I fear this may generate confusion... Not really sure of the alternatives though...

PR #1018, opened 2016-10-12
Fix issues regarding recurring events and DST

DanielBrierton:


web/public/js/controllers/dojo-event-form-controller.js
```diff
@@ -4,6 +4,10 @@
 
   function getEveryTargetWeekdayInDateRange(startDateTime, endDateTime, targetWeekday, eventType) {
     endDateTime = moment.utc(endDateTime).add(1, 'days');
+    var startTimeHour = moment(startDateTime.toDate()).hour();
+    var startTimeMin = moment(startDateTime.toDate()).minute();
+    var endTimeHour = moment(endDateTime.toDate()).hour();
+    var endTimeMin = moment(endDateTime.toDate()).minute();
```
Wardormeur: Fixed hours everybody :clap:

web/public/js/services/event-utils.js
```diff
@@ -51,8 +51,10 @@ angular.module('cpZenPlatform').factory('eventUtils', ['$translate', function($t
   eventUtils.getFormattedDates = function (event) {
     var utcOffset = moment().utcOffset();
 
-    var startDate = moment.utc(_.head(event.dates).startTime).subtract(utcOffset, 'minutes').toDate();
-    var endDate = moment.utc(_.head(event.dates).endTime).subtract(utcOffset, 'minutes').toDate();
+    var mStartDate = moment.utc(_.head(event.dates).startTime);
+    var mEndDate = moment.utc(_.head(event.dates).endTime);
+    var startDate = mStartDate.subtract(mStartDate.utcOffset(), 'minutes').toDate();
```
Wardormeur: Not sure to get the diff here

DanielBrierton: We're now getting the UTC offset from the date we're displaying and subtracting that instead of the UTC offset for the current date.

Wardormeur: Good catch

PR #1021, opened 2016-10-17
Feature/twilio integration

Wardormeur:


web/public/js/init-cdf.js
```diff
@@ -0,0 +1,302 @@
...
+
+  var failCb = function(err){
+    return {err: err};
+  };
+
+  var winCb = function(data){
+    return {data: data};
+  };
+
+  var resolves = {
```
DanielBrierton: We should probably centralise all our common resolves between init-master and init-cdf.

web/public/js/init-master.js
```diff
@@ -131,7 +131,11 @@
     },
     agreement: function(cdAgreementsService, $stateParams, $window, auth){
       return auth.get_loggedin_user_promise().then(function (user) {
-        return cdAgreementsService.loadUserAgreementPromise(user.id).then(winCb, failCb);
+        if (user) {
+          return cdAgreementsService.loadUserAgreementPromise(user.id).then(winCb, failCb);
+        } else {
+          winCb(void 0);
```
DanielBrierton: Shouldn't be a need for `void 0` here.

web/public/js/services/cd-api.js
```diff
@@ -1,46 +1,57 @@
...
+          method: 'DELETE',
+          url: baseUrl + url,
+          headers: {'Content-Type': 'application/json'}
+        })
+      }, resolve, reject);
    }
   };
+
+  function wrapCbPromise (fn, resolve, reject) {
+    if (resolve)
```
DanielBrierton: No curly braces make me sad 😢

web/public/js/directives/cd-expanding-card/index.js
```diff
@@ -19,9 +19,6 @@
       if (ctrl.startExpanded === 'true') {
         ctrl.expanded = 'expanded';
         content.css('height', 'auto');
-        requestAnimationFrame(function () {
```
DanielBrierton: Ok, yeah, I must have asked you to remove this.

PR #1023, opened 2016-10-19
Set cookie path so duplicates don't get set for CDF

DanielBrierton:
(specifically for angular-translate)

### DIFF
--- a/web/public/js/init-cdf.js
+++ b/web/public/js/init-cdf.js
@@ -228,8 +228,9 @@
         $httpProvider.interceptors.push('authHttpResponseInterceptor');
       }
     ])
-    .config(['$translateProvider',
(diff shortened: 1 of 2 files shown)


web/public/js/init-cdf.js
```diff
@@ -228,8 +228,9 @@
         $httpProvider.interceptors.push('authHttpResponseInterceptor');
       }
     ])
-    .config(['$translateProvider',
-      function($translateProvider) {
+    .config(['$translateProvider', '$cookiesProvider',
+      function($translateProvider, $cookiesProvider) {
+        $cookiesProvider.defaults.path = '/';
```
Wardormeur: because html base is different, the / reference /cdf/ instead of / ?

PR #1024, opened 2016-10-19
Take first translate cookie if multiple exist

DanielBrierton:


Wardormeur:
Do we need to introduce code for a scenario that "shouldn't" happen?

DanielBrierton:
I was thinking if we ever introduced another master js (maybe a standalone check-in web app) and we forgot the set the path or something like that. Not really sure it'd ever happen, but still, better safe than sorry?

PR #1027, opened 2016-10-20
Dojo list page can now be navigated to via url

eflanagan-r7:
Anchor links added for each country on the dojo list page

### DIFF
--- a/web/public/js/controllers/dojos-map-controller.js
+++ b/web/public/js/controllers/dojos-map-controller.js
@@ -130,7 +130,7 @@ function cdDojosMapCtrl($scope, $window, $state, $stateParams, $translate, $geol
   function loadDojosList() {
     // If we aren't on the list page already
     if ($stateParams.list === 'false' || !$stateParams.list) {
-      $state.go('home', { list: '' });
+      $state.go('home', { list: 'true' });
     }
 
     if(!$scope.dojoList || $scope.dojoList.length <= 0) {
--- a/web/public/js/controllers/dojos-map-controller.js
+++ b/web/public/js/controllers/dojos-map-controller.js
@@ -2,30 +2,14 @@
 /* global google,jQuery,MarkerClusterer */
 
 //  TODO : reuse cd-dojos-map instead of this mixed-up controller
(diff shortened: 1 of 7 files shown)


web/public/js/init-master.js
```diff
@@ -317,7 +318,8 @@
           templateUrl: '/dojos/template/dojo-list-index',
```
eflanagan-r7: This page seems to be broken as a result. Doesn't seem to be used though:
https://www.google.ie/?gws_rd=cr&ei=hesIWKnFM8nUgAbasrLwCA#q=site:%22http:%2F%2Fzen.coderdojo.com%22+intext:%22dojo-list-index%22

eflanagan-r7: Nvm. Seems to be already broken

DanielBrierton:
Addresses #community-platform/829

web/public/js/controllers/dojos-map-controller.js
```diff
@@ -105,7 +125,14 @@ function cdDojosMapCtrl($scope, $window, $state, $stateParams, $translate, $geol
     });
   });
 
-  $scope.loadDojosList = function(){
+  $scope.loadDojosList = loadDojosList;
+
+  function loadDojosList() {
+    // If we aren't on the list page already
+    if ($stateParams.list === 'false' || !$stateParams.list) {
+      $state.go('home', { list: '' });
```
DanielBrierton: This needs a value for the URL to change when clicking the tab. You had it working at one point, but this change breaks it. :(

web/public/js/directives/cd-anchor/index.js
```diff
@@ -0,0 +1,23 @@
+;(function() {
```
DanielBrierton: Minor thing, but we use 2 spaces for indentation. Since you need to fix the other comment, it'd be nice if you could fix the indentation here too.

web/public/js/directives/cd-anchor/template.dust
```diff
@@ -0,0 +1,4 @@
+<div>
```
DanielBrierton: Same with this, re: indentation.

eflanagan-r7: Hey I think these are okay now. Any other problems let me know

DanielBrierton:
Sorry to keep you waiting, we've had a busy weekend with DojoCon, and I'm only just getting back into the swing of things now. Thanks again for the pull request. All looks good now, so I'm going to merge it.

PR #1033, opened 2016-11-02
Fix unhandled download export links

Wardormeur:
remove front-end treatment which was having browser compat issues
replace with an API call returning CSV with the right data type


DanielBrierton:
❤

PR #1037, opened 2016-11-07
Remove cors for api calls not requiring auth

Wardormeur:


lib/dojos.js
```diff
@@ -840,6 +846,7 @@ exports.register = function (server, options, next) {
       description: 'uncompleted',
       notes: 'uncompleted',
       tags: ['api', 'dojos'],
+      cors: { origin: ['*'], credentials: false },
```
DanielBrierton: Remove cors, use actHandlerNeedsUser

PR #1044, opened 2016-11-17
Bugfix/check deps

Wardormeur:
Fix multiple lib version
Remove unused dep
Clean DOM & js from removed dep


web/public/dependencies.json
```diff
@@ -56,7 +55,6 @@
   "./web/public/components/ng-idle/angular-idle.min.js",
   "./web/public/components/angular-dynamic-locale/tmhDynamicLocale.min.js",
   "./web/public/components/angularjs-dropdown-multiselect/dist/angularjs-dropdown-multiselect.min.js",
-  "./web/public/components/angular-tooltips/dist/angular-tooltips.min.js",
```
DanielBrierton: Do we not use this? If not, why not remove it from bower.json as well?

PR #1046, opened 2016-11-26
Cancel/Dismiss event

Wardormeur:
Interested in knowing what we want for "save draft". Should we allow a published event to go back to draft?
Can one cancel a draft event or should it be deleted instead? @rosalanghammer @DanielBrierton 
Also, fix "Use dojo location" when using a draft, the dojo was never recoverd, hence impossible to use its location.

### DIFF
--- a/web/public/js/controllers/dojo-event-form-controller.js
+++ b/web/public/js/controllers/dojo-event-form-controller.js
@@ -279,11 +279,23 @@
 
     $scope.eventInfo.invites = [];
 
-    $scope.cancel = function($event) {
+    $scope.dismiss = function($event) {
       $event.preventDefault();
       $event.stopPropagation();
       deleteLocalStorage();
       goToManageDojoEvents($state, null, dojoId);
+    };
+
+    $scope.cancel = function($event) {
+      $event.preventDefault();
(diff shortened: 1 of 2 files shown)


rosalanghammer:
Draft events should be 'Discarded' i.e. function is delete not cancel as otherwise it clutters up the manage event page. 
 
Then baring this in mind no published event should be allowed to go back to draft. Otherwise we could give users the permission to delete events without informing people who are signed up.

Wardormeur:
Dismiss changes has beenmerged through another PR, need to be rebased to support deletion

web/public/js/controllers/dojo-event-form-controller.js
```diff
@@ -279,13 +279,25 @@
...
 
-    $scope.cancel = function($event) {
+    $scope.dismiss = function($event) {
       $event.preventDefault();
       $event.stopPropagation();
       deleteLocalStorage();
       goToManageDojoEvents($state, null, dojoId);
     };
 
+    $scope.cancel = function($event) {
```
DanielBrierton: Should have a confirmation before cancelling the event

PR #1051, opened 2016-12-02
Created a key joining "I have read and accept the" and "Terms and Conditions here" to resolve part of issue #934

rhemon:
https://github.com/CoderDojo/community-platform/issues/934

The issue mentioned that this two part should have been joined to resolving translating errors, and so I have tried forming a key joining both together. Visually nothing seemed to have changed, but locally I couldn’t test translation.

### DIFF
--- a/web/locale/en_US/messages.po
+++ b/web/locale/en_US/messages.po
@@ -3154,3 +3154,6 @@ msgstr "You have been invited to a Dojo session"
 
 msgid "Change Role"
 msgstr "Change Role"
+
+msgid "I have read and accept the Terms and Conditions here"
+msgstr "I have read and accept the <a ui-sref='terms-and-conditions' target='_blank'><b>Terms and Conditions here</b></a>"
--- a/web/locale/en_US/messages.po
+++ b/web/locale/en_US/messages.po
@@ -3155,5 +3155,5 @@ msgstr "You have been invited to a Dojo session"
 msgid "Change Role"
 msgstr "Change Role"
 
-msgid "I have read and accept the Terms and Conditions here"
+msgid "I have read and accept the <a ui-sref=\"terms-and-conditions\" target=\"_blank\"><b>Terms and Conditions here</b></a>"
 msgstr "I have read and accept the <a ui-sref=\"terms-and-conditions\" target=\"_blank\"><b>Terms and Conditions here</b></a>"
--- a/web/locale/en_US/messages.po
+++ b/web/locale/en_US/messages.po
@@ -3156,4 +3156,4 @@ msgid "Change Role"
 msgstr "Change Role"
 
 msgid "I have read and accept the Terms and Conditions here"
-msgstr "I have read and accept the <a ui-sref='terms-and-conditions' target='_blank'><b>Terms and Conditions here</b></a>"
+msgstr "I have read and accept the <a ui-sref=\"terms-and-conditions\" target=\"_blank\"><b>Terms and Conditions here</b></a>"
--- a/web/public/js/directives/user/cd-register-user/template.dust
+++ b/web/public/js/directives/user/cd-register-user/template.dust
@@ -38,8 +38,7 @@
 
       <div class="checkbox cd-checkbox checkbox-primary">
         <input type="checkbox" name="termsConditionsAccepted" ng-class="{'has-error': registerForm.$submitted && registerForm.termsConditionsAccepted.$error.required}" ng-model="formData.user.termsConditionsAccepted" required required-message="'{@i18n key="You must accept the Terms and Conditions"/}'" id="termsAndConditionsCheckbox">
-        <label for="termsAndConditionsCheckbox">{@i18n key="I have read and accept the"/}
-          <a ui-sref="terms-and-conditions" target="_blank"><b>{@i18n key="Terms and Conditions here"/}</b></a>
+        <label for="termsAndConditionsCheckbox">{@i18n key="I have read and accept the Terms and Conditions here"/}
         </label>
       </div>
     </div>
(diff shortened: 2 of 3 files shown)


rhemon:
I tried testing with messages.po, the text changed on localhost, so according to me there should be no mistake in functionality of translation

Wardormeur:
What 's your thinking @niccokunzmann ?

niccokunzmann:
@Wardormeur what is your process: Do you merge optimistically or do you like to merge after critique?

I would like to know the following and we can test it out:
The message ID and the message differ. What turns up in crowdin? The message ID or the English translation?
If it is the English translation, then fine. If it is the message ID, other people will not know of the link.

This is a possibility to try this out and if the Message ID is taken, we can create another issue and fix it with a second PR.

So, my questions: 

1. Do you know, what Crowdin uses?
2. Is optimistic merging ok? http://hintjens.com/blog:106

niccokunzmann:
@rhemon I think it may be important to have the double quotes:

    "terms-and-conditions" target="_blank"

instead of

    'terms-and-conditions' target='_blank'

rhemon:
alright I will do those change just a second, btw could you explain me the role of translation-key.js file, like why are you doing that and is messages.po connected?

niccokunzmann:
Maybe @Wardormeur knows but @DanielBrierton should also know the role of the translations file. (Source: [git blame](https://github.com/rhemon/cp-zen-platform/blame/00f195054bf92a674f5c7c1f733723e9c8d8436b/web/public/js/services/translation-keys.js))

Wardormeur:
Most of the time, if it's a modification like this, we're going optimistic :)
If it's a huge rewrite (5+ files), we prefer to discuss it/review it more deeply
Crowdin uses the key to "regroup" translations, whereas we use the english translation when none exists in other languages (when using the cp-i18n, not sure w/ angular) . We haven't found a way to "change a key" thought, even less notify users of a change.
In this case, it should be a match of the key AND the value, as Crowdin only shows up the key in their interface, meaning most translation have a risk to NOT contain the link since we don't expect people to compare the value to others strings (see the case "error.general" which stayed the same whereas it has a translation in English).
The role of translations-keys is to be able, on the front-end to define some keys which are immutable and technical oriented (like error.general :)) without having an impact on Crowdin. It's an interface for development to avoid having to use the whole string, but a keyword instead. It's only used when using $translate from angular atm, but should probably be expanded to cp-i18n.

niccokunzmann:
As an addition: The case error.general, look at the languages tab after login: https://crowdin.com/translate/zen-community-platform/1/en-de#q=error.general most people do not look at the English text but at the message ID.

rhemon:
btw the linking works fine with the existing code without using double quotes

rhemon:
secondly when I try adding anchor tag in msgid, it doesn't replace with the string in msgstr

niccokunzmann:
@rhemon to double vs single quote: Stackoverflow backs your view:  http://stackoverflow.com/questions/2373074/single-vs-double-quotes-vs

niccokunzmann:
@rhemon Can you push the changes which do not replace to some branch so we can have a look at it?

rhemon:
i added the double-quotes, ahd here is a separate branch https://github.com/rhemon/cp-zen-platform/tree/another-branch

rhemon:
is msgid something that should just finds the exact same string and replaces it with the msgstr? is that why if I add anchor tag to msgid the link part doesn't work?

niccokunzmann:
I commented: https://github.com/rhemon/cp-zen-platform/commit/cadf30e2070d90f5d2d6bcef75761697b9628d1f#commitcomment-20053001

niccokunzmann:
To your question: There are [messageids with a link inside](https://github.com/rhemon/cp-zen-platform/blob/00f195054bf92a674f5c7c1f733723e9c8d8436b/web/locale/en_US/messages.po#L243). Having a link should work. But if only one character is off, it is not found any more.

rhemon:
yeah seen the comment, i figured out its not a problem in msgid or msgstr
but in the translation-key.js, when I remove both msgid and msgstr, the link thing doesnt work, that means the default one taken from key is the one with no link

niccokunzmann:
@rhemon What I understood: with the message id it is okay, if the link is inside but somehow translation-key.js is used and changes the outcome. 
I do not understand the role of translation-key.js.

rhemon:
i am gonna try manually adding it to the template without the usage of translation-key.js

rhemon:
since previously the anchor tag was manually added in template, but string were brought from the js file

rhemon:
yes fixed it

rhemon:
what I had to do was make the key name same, that way it got recognised in msgid and replaces it with msgstr. It worked for me @niccokunzmann @Wardormeur

rhemon:
crowdin here is also updated properly : https://crowdin.com/translate/zen-community-platform/1/en-de#q=i+have+read+and+

niccokunzmann:
Cool!

niccokunzmann:
So, we can merge and I can translate it to German.

rhemon:
I feel like this was one of the best accomplishment in GCI :) learnt quite a lot

niccokunzmann:
I also learned something.. 

    Ich habe die <a ui-sref="terms-and-conditions" target="_blank">allgemeinen Geschäftsbedingungen</b></a> gelesen und akzeptiere sie.

It is translated.

niccokunzmann:
@Wardormeur I think, you can merge, crowdin builds and we see the German translation live.

niccokunzmann:
If there is still an issue, we can create a new issue and fix it much faster.

Wardormeur:
GJ !

rhemon:
I hope there wont be any issue

niccokunzmann:
@Wardormeur Do you know when it will be live?

DanielBrierton:
@niccokunzmann Code is automatically deployed when we merge to master, so it should be live now. Thanks for the PR @rhemon!

PR #1053, opened 2016-12-03
Fix CoderDojo/community-platform#1118

niccokunzmann:
I changed the key which is looked up the the key used in the messages.po file.
Also see #1051 and [this](https://github.com/niccokunzmann/cp-zen-platform/pull/1) for the fun of it.
Issue: CoderDojo/community-platform#1118


Wardormeur:
Will merge tomorrow morning if @DanielBrierton doesn't have any comment :)

PR #1054, opened 2016-12-04
Fixed the header and footer Translation erros mentioned in #934

rhemon:
The translation that weren’t added on header and footer, I have fixed those. @niccokunzmann @Wardormeur 
Please check


rhemon:
there are too many commits so I am gonna try squashing it and the do the pull request again

PR #1060, opened 2016-12-12
Updated circle.yml to only run crowdin on master, srcclr moved to circle project conf

DanielBrierton:


butlerx:
is it not a good idea to run srcclr on prs?

PR #1064, opened 2016-12-17
Mobilepic

all-self-taught:
Updated the CSS to allow the mobile pic for issue #1123. Also modified the HTML to account for new classes.


package.json
```diff
@@ -26,6 +26,7 @@
   },
   "dependencies": {
     "async": "0.9.2",
+    "babel-core": "^6.21.0",
```
butlerx: wheres this used?

web/public/css/profile-image.less
```diff
@@ -0,0 +1,23 @@
+@media (max-width:767px){
+	.mobile-center {
+		margin: auto;
+		width: 50%;
+		text-align: center;
+	}
+	.img-responsive{
```
butlerx: overriding the bootstrap class could break other images

all-self-taught:
babel-core isn't really used, I was having problems getting my commit to pass the test, so I was googling around for answers and must've installed it. It turned out the issue was I was using the wrong node version, so I downgraded node, and it worked.

How can I verify that overriding the bootstrap class will not break the other images?

butlerx:
only way to verify is check by hand, but we previously had problems of overrides doing unexpected things in random places. Your best bet is to create a class such as .cd-img-responsive that does the override and add it to where you want to override. 
unless @DanielBrierton or  @Wardormeur would like it done a different way?

web/public/css/profile-image.less
```diff
@@ -0,0 +1,23 @@
+@media (max-width:767px){
+	.mobile-center {
+		margin: auto;
+		width: 50%;
```
DanielBrierton: I get the thought process, but the cd-round-image has a width of 200px, which can overflow this 50% when at resolutions like 320px (iPhone 4/S, 5/S, SE all have this width). 200px is a nice size on mobile anyway, so I'd recommend just removing this width, and having the 200px width stay as is.

DanielBrierton:
Fixed since by #1092, so closing.

PR #1073, opened 2017-01-13
Bugfix/allow profile visibility

Wardormeur:


web/public/js/controllers/user-profile-controller.js
```diff
@@ -235,13 +236,14 @@ function cdUserProfileCtrl($scope, $rootScope, $state, $window, auth, cdUsersSer
...
     });
 
-    cdUsersService.loadDojoAdminsForUserPromise(loggedInUser.data.id).then(
-      function(champions){
-        $scope.loggedInUserIsMemberOfDojoChampion = !!(_.find(champions, function (championForUser) {
-          return championForUser.id === profileUserId;
-        }));
-      }
-    );
+    cdUsersService.loadChildrenForUser($stateParams.userId, function (children) {
```
DanielBrierton: Should these check the user type first? Make sure their an adult before loading children, and a child before loading parents. Nothing major really, just a minor performance improvement by freeing up a connection for a bit. Could also check `$scope.profile.children.length` to see if there's any point fetching.

web/public/js/controllers/user-profile-controller.js
```diff
@@ -523,24 +525,24 @@ function cdUserProfileCtrl($scope, $rootScope, $state, $window, auth, cdUsersSer
...
         if(loggedInUserIsDojoAdmin()) return true;
         return !$scope.isPrivate;
       case 'parent-guardian':
-        if($scope.profile.ownProfileFlag) return true;
+        if($scope.ownProfileFlag) return true;
         if(loggedInUserIsChampion()) return true;
         if(loggedInUserIsDojoAdmin()) return true;
         if(loggedInUserIsChild()) return true;
-        return false; //Always private
+        return !$scope.isPrivate; //Always private
```
DanielBrierton: No longer `//Always private`

web/public/js/controllers/user-profile-controller.js
```diff
@@ -235,13 +236,14 @@ function cdUserProfileCtrl($scope, $rootScope, $state, $window, auth, cdUsersSer
...
-        $scope.loggedInUserIsMemberOfDojoChampion = !!(_.find(champions, function (championForUser) {
-          return championForUser.id === profileUserId;
-        }));
-      }
-    );
+    cdUsersService.loadChildrenForUser($stateParams.userId, function (children) {
+      $scope.profile.resolvedChildren = children;
+    });
+
+    cdUsersService.loadParentsForUserPromise($stateParams.userId)
```
DanielBrierton: The resolve for parentsForUser is still being used along with this.

Same point as well as regarding `$scope.profile.parents.length`.

I'd also like to see a loading animation in these sections if we need to fetch children or parents.

Also, this uses a Promise, children uses a callback. Probably should use Promise for children as well for consistency.

web/public/js/controllers/user-profile-controller.js
```diff
@@ -576,7 +579,7 @@ function cdUserProfileCtrl($scope, $rootScope, $state, $window, auth, cdUsersSer
   function loggedInUserIsParent() {
     if(!loggedInUser.data) return false;
     return _.find(parentsForUser.data, function (parentForUser) {
```
DanielBrierton: Probably drop using this resolve and just profile.data.parents just like is done for loggedInUserIsChild

web/public/js/directives/user/cd-profile/view/template.dust
```diff
@@ -37,7 +37,8 @@
             <badges-list ng-if="!hideProfileBlock('badges')"></badges-list>
             <projects-list ng-if="!hideChampionProfileBlock('projects')"></projects-list>
             <parents-list ng-if="!hideProfileBlock()"></parents-list>
-            <children-list ng-if="loggedInUserIsCDFAdmin() || hasAccess(profile.userTypes, ['parent-guardian', 'mentor', 'champion'])"></children-list>
+            <!-- ng-if="loggedInUserIsCDFAdmin() || hasAccess(profile.userTypes, ['parent-guardian', 'mentor', 'champion'])"  -->
```
DanielBrierton: Remove

PR #1076, opened 2017-01-18
Remove hiding of 400 errors, and fix Joi for create dojo

DanielBrierton:
This time with a version of Boom that works on node 0.10!


Wardormeur:
:boom:

PR #1081, opened 2017-01-24
Disallow creation of event without datesµ

Wardormeur:


web/locale/en_US/messages.po
```diff
@@ -3232,3 +3232,6 @@ msgstr "Click one of your Dojos to add {{ name }} as a member, or <a href=\"{{ f
 
 msgid "We could not determine a location from this address. Your previous location marker has been retained."
 msgstr "We could not determine a location from this address. Your previous location marker has been retained."
+
+msgid "Day selected doesn\'t exist in date range, please check your date range."
```
DanielBrierton: Is the backslash needed here? Did you test to be sure `$translate.instant` works and isn't just returning the key as the translation?

PR #1019, opened 2016-10-13
add first version languages images issue #1043

benjamingarcia:
That's ok? [issue #1043](https://github.com/CoderDojo/community-platform/issues/1043)

PR #1083, opened 2017-01-25
Apply a filter on activity for dojo map search results

Wardormeur:
Optimatly, it should be done b-end side

### DIFF
--- a/web/public/js/controllers/dojos-map-controller.js
+++ b/web/public/js/controllers/dojos-map-controller.js
@@ -48,10 +48,8 @@ function cdDojosMapCtrl($scope, $window, $state, $stateParams, $translate, $geol
         deleted: 0,
         fields$: ['name', 'geo_point', 'stage', 'url_slug', 'private']
       }, function (dojos) {
-        var filteredDojos = [];
-        _.each(dojos, function (dojo) {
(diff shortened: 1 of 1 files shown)


Wardormeur:
closes #1019

web/public/js/controllers/dojos-map-controller.js
```diff
@@ -211,9 +209,18 @@ function cdDojosMapCtrl($scope, $window, $state, $stateParams, $translate, $geol
     $scope.markers = [];
   }
 
+  function filterDojos (dojos) {
```
DanielBrierton: `filterInactiveDojos` would be more descriptive of what it actually does

web/public/js/controllers/dojos-map-controller.js
```diff
@@ -211,9 +209,18 @@ function cdDojosMapCtrl($scope, $window, $state, $stateParams, $translate, $geol
...
+    var filteredDojos = [];
```
DanielBrierton:
```javascript
return _.filter(dojos, function (dojo) {
  return dojo.stage !== 4;
});
```

Wardormeur: Creppy pasta :>

PR #1099, opened 2017-02-28
changed recurring to term issue in events ticketing

BSKERRITT:
https://github.com/CoderDojo/community-platform/issues/1067

### DIFF
--- a/web/locale/en_US/messages.po
+++ b/web/locale/en_US/messages.po
@@ -1605,10 +1605,10 @@ msgstr "Mentor/Volunteer"
 msgid "Champion"
 msgstr "Champion"
 
-msgid "recurring"
-msgstr "Recurring"
+msgid "Term"
+msgstr "Term"
 
-msgid "one-off"
+msgid "One off"
 msgstr "One off"
 
 msgid "Select event date..."
--- a/web/locale/en_US/messages.po
+++ b/web/locale/en_US/messages.po
@@ -1608,9 +1608,6 @@ msgstr "Champion"
 msgid "Term"
 msgstr "Term"
 
-msgid "One off"
-msgstr "One off"
-
 msgid "Select event date..."
 msgstr "Select event date..."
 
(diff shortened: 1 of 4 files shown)


web/locale/en_US/messages.po
```diff
@@ -1605,10 +1605,10 @@ msgstr "Mentor/Volunteer"
 msgid "Champion"
 msgstr "Champion"
 
-msgid "recurring"
-msgstr "Recurring"
+msgid "Term"
+msgstr "Term"
 
-msgid "one-off"
+msgid "One off"
```
Wardormeur: Loss of existing translations for no benefit

web/public/templates/dojos/events/user-events.dust
```diff
@@ -24,7 +24,7 @@
             <tr ng-click="showEventInfo(dojoEvents, $index)" class="event-info-row">
               <td>{{ event.name }}</td>
               <td>{{ event.formattedDate }}</td>
-              <td>{{ event.type }}</td>
+              <td>{{ getEventTypeDisplayText(event) }}</td>
```
Wardormeur: Double binding on a function returning a translation, largely overkill, prefer storing it in a var and binding once

web/locale/en_US/messages.po
DanielBrierton: Actually turns out there's already a "One off" msgid lower down in the file, so @BSKERRITT is removing this one. one-off as a msgid was actually never used.

web/public/templates/dojos/events/user-events.dust
DanielBrierton: So, like
```javascript
_.each(dojoEvents.events, function(event){
  event.translatedType = getEventTypeDisplayText(event);
});
```
?

Wardormeur: Didn't see it was a ng-repeat, hence the use of a function can be justified while there is no existing, preparing, loop like you have describe. Even better, there is only 2 possibles outputs, so that an key-val object could be used to store the value to avoid consecutive calls to ng-translate 
Finally my remark on binding is that this tpl is display oriented, hence it doesn't make sense to double bind since the model shouldn't change

DanielBrierton: Actually, there's a loop on the events already, buried deep inside `user-events-controller.js`. key-val object also makes sense. I still prefer the use of a function over adding an extra property to the event objection, but yeah, one-time binding makes sense here. On all the cells really.

PR #1102, opened 2017-03-13
Fix upgrade of an dojo-user if there is no existing perm

Wardormeur:


Wardormeur:
ref https://forums.coderdojo.com/topic/339/changing-role-to-champion-not-working/2

PR #1103, opened 2017-03-15
add visibility of parent/kid connection on manage-dojo-users

Wardormeur:
Fixes https://github.com/CoderDojo/community-platform/issues/1076

### DIFF
--- a/web/public/js/directives/cd-picture-grid/style.less
+++ b/web/public/js/directives/cd-picture-grid/style.less
@@ -18,6 +18,10 @@
     &:hover {
       text-decoration: none;
 
+      .cd-picture-grid__picture-sub {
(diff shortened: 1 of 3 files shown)


DanielBrierton:
Need to look into why the image doesn't load.

Also, a question for @rosalanghammer on this.
* Parent A has Child A
* Parent A and Child A are a member of Dojo A
* Parent A adds Parent B to Child A
* Does Parent B automatically get added to Dojo A when they accept the invite?

Wardormeur:
bump @rosalanghammer  :)

rosalanghammer:
Yes automatically added to all Dojos their Child is part of. @DanielBrierton @Wardormeur

PR #1115, opened 2017-04-11
Change srcFallback handling or rounded-image

Wardormeur:
http head if done too soon isn't passing down every info to respect cors?
Fix https://github.com/CoderDojo/community-platform/issues/1179

### DIFF
--- a/web/public/js/directives/cd-rounded-image/index.js
+++ b/web/public/js/directives/cd-rounded-image/index.js
@@ -15,18 +15,17 @@ function cdRoundedImage(){
         alt: '@',
         size: '=?',
         edit: '=?',
-        upload: '=?',
-        format: '=?' // fallback for base64 to avoid the S3 request
+        upload: '=?'
       },
       restrict: 'EA',
       templateUrl: '/directives/tpl/cd-rounded-image',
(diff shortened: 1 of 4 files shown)


web/public/js/directives/cd-rounded-image/index.js
```diff
@@ -40,16 +39,16 @@ function cdRoundedImage(){
...
-            idWatcher();
-            if(_.isUndefined(cdRI.format)) {
-              //  TODO: use a provider like cd-Entity-Service (i.e: cd-dojos-service=>getAvatar) to get the proper image
-              $http.head(cdRI.src).then(function successCallback(response) {
-                //Nothing to do, it all works as expected, the image is stored online
-              }, function errorCallback(response) {
-                //File doesn't exists, we should fallback
+            $timeout(function () {
+              var image = angular.element(
+                document.getElementsByClassName(cdRI.edit ?
```
DanielBrierton: Potential to trigger on error of a different image if directive is used multiple times on one page.

Inject `$element` and use `$element.find` instead of `document.getElementsByClassName`. Limits the search to this directive's template only.

PR #1128, opened 2017-05-04
Enhancement/SADv2

Wardormeur:
WIP don't merge


Wardormeur:
Will go through staging-> master

PR #1130, opened 2017-05-08
New url to redirect to edit profile without profile id

grabartley:
[#1030](https://github.com/CoderDojo/community-platform/issues/1030)

### DIFF
--- a/web/public/js/init-master.js
+++ b/web/public/js/init-master.js
@@ -757,16 +757,12 @@
         //redirect to edit-user-profile page without need for userId in url
         .state("edit-my-profile", {
           url: '/profile/edit',
-          controller: function ($state, auth) {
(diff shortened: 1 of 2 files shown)


web/public/js/init-master.js
```diff
@@ -754,6 +754,15 @@
           },
           templateUrl: '/directives/tpl/user/cd-profile/edit'
         })
+        //redirect to edit-user-profile page without need for userId in url
+        .state("edit-my-profile", {
+          url: '/profile/edit',
+          controller: function ($state, auth) {
+            auth.get_loggedin_user_promise().then(function(user) {
```
Wardormeur: What if not logged-in?
Maybe make it part of parent:dashboard too so that it'll redirect if that's not the case?

PR #1093, opened 2017-02-16
Added add child page

DanielBrierton:
Also improves usage of firstName/lastName and datepickers for DoB.

PR #1131, opened 2017-05-08
Cookie policy is now accepted on navigation

grabartley:
[#1093](https://github.com/CoderDojo/community-platform/issues/1093)

### DIFF
--- a/web/public/js/init-master.js
+++ b/web/public/js/init-master.js
@@ -1,5 +1,6 @@
 (function() {
   'use strict';
+  /* global jQuery */
 
   var gmap = function($q, $window) {
     var dfd = $q.defer();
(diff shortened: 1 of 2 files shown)


web/public/js/init-master.js
```diff
@@ -1045,12 +1046,35 @@
         }
         pageTitle.push("CoderDojo Zen");
         $rootScope.pageTitle = pageTitle.join(" | ");
+
+        //if moving from another page
+        if (!fromState.abstract) {
```
Wardormeur: abstract means it has a superState to which you can't navigate : it should always be false, basically.
Instead, maybe try $cookie.get('cookieDisclaimer') to verify it's existing or not
Elsewhat i'll be forever refreshed, meaning the cookie TTL will be.. well, non-respected.

PR #1136, opened 2017-05-12
Add max image size to event descriptions

howardjing:
Currently event description images can be added at any size. This
adds max-width / max-height to prevent images from being larger
than their parents.

Addresses CoderDojo/community-platform#1028

Before:
image
After:
image


Wardormeur:
Very nice, thanks :)

PR #1138, opened 2017-05-14
added joi annotations to the agreements api.

meetjonblack:
Taking a crack at [Swagger docs for the full API](https://github.com/CoderDojo/community-platform/issues/750)

Updated Agreements api


lib/agreements.js
```diff
@@ -1,42 +1,97 @@
...
+        'hapi-swagger': {
+          responseMessages: [
+            {code: 400, message: 'Bad Request'},
+            {code: 200, message: 'OK'}
+          ]
+        }
+      },
+      validate: {
+        payload: Joi.object({
+          agreement: {
```
Wardormeur: This is too restrctive, the payload used to save an agreement is 
{ agreementVersion: 2
  fullName: "derp"
  userId: "8735c503-bd3a-433f-9337-afd789cb7fec" }

lib/agreements.js
```diff
@@ -1,42 +1,97 @@
...
+            userId: Joi.alternatives().try(Joi.string().guid(), Joi.string()).required()
+          }
+        })
+      }
     }
   }, {
     method: 'POST',
     path: options.basePath + '/agreements/list-by-ids',
-    handler: handlers.actHandler('get_agreements'),
+    handler: handlers.actHandler('list'),
```
Wardormeur: if list-by-ids/get_agreement (which doesn't exists?) is not used, i'd prefer it to be removed rather than replaced

lib/agreements.js
```diff
@@ -1,42 +1,97 @@
...
+      plugins: {
+        'hapi-swagger': {
+          responseMessages: [
+            {code: 400, message: 'Bad Request'},
+            {code: 200, message: 'OK'}
+          ]
+        }
+      },
+      validate: {
+        payload: Joi.object({ query: {
```
Wardormeur: This is used by the stats page in /dashboard/stats (login as manager1).
The current payload used is {query : {agreement_version: 2}} because we mostly care about the latest version of the agreement :)

lib/agreements.js
```diff
@@ -1,42 +1,97 @@
...
+          responseMessages: [
+            {code: 400, message: 'Bad Request'},
+            {code: 200, message: 'OK'}
+          ]
+        }
+      },
+      validate: {
+        payload: Joi.object({
+          agreement: {
+            userId: Joi.alternatives().try(Joi.string().guid(), Joi.string()).required()
```
Wardormeur: guid is actually more correct than string, thanks for that :+1:

lib/agreements.js
```diff
@@ -1,42 +1,97 @@
...
+      }
     }
   }, {
     method: 'GET',
     path: options.basePath + '/agreements/{id}',
     handler: handlers.actHandler('load_user_agreement', 'id'),
     config: {
-      description: 'Load user agreement',
-      tags: ['api', 'users']
+      description: 'Load a user agreement by id',
```
Wardormeur: by his user_id
It's "id" which is passed down, but it's actually a user_id :(

meetjonblack:
I see what you mean. I am unsure as to what you mean by checking out the db structure. I couldn't find an agreements table.  Where would I look exactly to determine this or any other api?

Wardormeur:
The table for that is in the database called cp-users-development, table cd_agreements
When it comes to an API.. well, those endpoint are not properly documented for a reason : it takes some time to figure out the scenario once the API has been implemented :( Generally I look over the cp-e2e tests to see if they are covered and run the test posts changes of the API restriction, elsewhat I look for usage of this API and look if I may have been overlooking fields/usages.
Most of all, for /list/search functions, refer to https://github.com/senecajs-labs/seneca-store-query

package.json
```diff
@@ -43,12 +43,12 @@
     "good": "6.3.0",
     "good-file": "5.0.1",
     "good-http": "5.0.x",
-    "hapi": "10.4.1",
+    "hapi": "14.2.0",
```
Wardormeur: Don't.
There is so many changes to cover, it requires its own update.

lib/profiles.js
```diff
@@ -5,185 +5,241 @@ var cacheTimes = require('../web/config/cache-times');
 var fs = require('fs');
 var defaultImage = new Buffer(fs.readFileSync(__dirname + '/../web/public/img/avatars/avatar.png', 'base64'), 'base64');
 var auth = require('./authentications');
+var Joi = require('joi');
+var ZenJoi = require('./zenJoi');
```
Wardormeur: I invite you to do it as a second PR to facilitate review

ArayB:
Closing as this is very old.

PR #1154, opened 2017-07-18
moved text right in alert

all-self-taught:
Placed text in red alert and moved to the right.

### DIFF
--- a/web/public/css/_variables.less
+++ b/web/public/css/_variables.less
@@ -30,4 +30,4 @@
 @state-warning-bg: @cd-yellow;
 @state-warning-text: @cd-gray;
 @state-danger-bg: @cd-red;
-@state-danger-text: @cd-red;
+@state-danger-text: @cd-white;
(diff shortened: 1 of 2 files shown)


web/public/templates/dojos/events/user-events.dust
```diff
@@ -3,42 +3,45 @@
...
-              <td colspan="9">
-                <div class="span9">
-                  <cd-apply-for-event></cd-apply-for-event>
-                </div>
-              </td>
-            </tr>
-            </tbody>
-          </table>
-        </div>
+  <!--<div class="panel panel-body" ng-if="!dojosEvents || dojosEvents.length === 0 || !currentEvents">
```
Wardormeur: Remove comment? :)

web/public/css/_variables.less
```diff
@@ -30,4 +30,4 @@
 @state-warning-bg: @cd-yellow;
 @state-warning-text: @cd-gray;
 @state-danger-bg: @cd-red;
-@state-danger-text: @cd-red;
+@state-danger-text: @cd-white;
```
Wardormeur: That's a bootstrap-wide change, do you see any scenario where we override the background on danger ?@DanielBrierton

all-self-taught: I did a grep on the danger alert, and found no other results.

web/public/templates/dojos/events/user-events.dust
all-self-taught: Comment was left in there in case you simply wanted the text moved, as opposed to a red alert.

Wardormeur:
This page has been reworked in the meantime, closing off, sorry @all-self-taught

PR #1157, opened 2017-07-21
Fix frontend word wrap in dojo's sidebar-contact

rstein66:
Improve readability of sidebar contact info on dojo's main page by specifying that lines should break between words (`break-word`), instead of between any 2 chars (`break-all`).

Closes CoderDojo/community-platform/issues/1207


Wardormeur:
Please don't introduce a package-lock.
We don't have one for a simplicity reason (internal dependency which is trans-repos, yet need to be updated on each repo : hence master is always picked).
Elsewhat, thanks for the PR :+1:

rstein66:
Oops & thanks! I've removed the file

Wardormeur:
We've released our new front-end (including the dojo page), hence this page isn't used anymore. 
We'll have a proper annoucement soon so that we clearly don't have contribution on dead pieces of code.
Apologies for that :/

rstein66:
I understand; thanks for the update!

PR #1161, opened 2017-07-31
Enhancement/staging SADv2 31 07 2017

Wardormeur:


lib/dojos.js
```diff
@@ -359,6 +359,7 @@ exports.register = function (server, options, next) {
       cors: { origin: ['*'], credentials: false },
       validate: {
         payload: Joi.alternatives().try(Joi.any(), Joi.object({ query: {
+          id: Joi.alternatives(Joi.string().guid(), Joi.object()),
```
Wardormeur: allows us to use nin$

PR #1170, opened 2017-08-14
#150170737 Fix error handling for twitter/facebook for dojo & profile

Wardormeur:


Wardormeur:
Seem to introduce a(nother) bug when a "dot" is typed in email field (ie registration)
A fork of ng-validator is probably required to only take what we want out of the diff 1.2.9-1.2.10

PR #1179, opened 2017-09-02
Upgrade/eslint

butlerx:
Move all configs to the root of the repo
replace  semistandard with eslint
remove unused files such as express config and olf build coscript


index.js
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env node
+const TITLE = 'cp-zen-platform';
```
Wardormeur: Isn't this pointless considering our arch now?

web/index.js
```diff
@@ -1,403 +1,408 @@
...
+const eventbrite = require('../lib/eventbrite.js');
+const polls = require('../lib/polls.js');
+const orgs = require('../lib/organisations.js');
+const dojos = require('../lib/dojos.js');
+const dustI18n = require('./lib/dust-i18n.js');
+const dustLoadJs = require('./lib/dust-loadjs.js');
+const dustLoadOG = require('./lib/dust-load-open-graph.js');
+
+exports.start = () => {
+  dustI18n();
```
Wardormeur: no need for those ?

web/index.js
```diff
@@ -1,403 +1,408 @@
...
-  }
-}, function (err) {
-  checkHapiPluginError('cp-zen-frontend')(err);
-});
-
-server.ext('onPreAuth', function (request, reply) {
-  var translateCookie = request.state && request.state.NG_TRANSLATE_LANG_KEY;
-  if (_.isArray(translateCookie)) {
-    translateCookie = translateCookie[0];
+  if (env === 'production' || env === 'staging') {
```
Wardormeur: this is useless, handled by loadB

butlerx:
eslint ignores web/public and leaves jshint to deal with it as it always has

web/index.js
butlerx: if you look at the files they were modified to be functions but i can revert that if need be

butlerx:
closing in favour of #1189

PR #1180, opened 2017-09-02
migrate to circlev2

butlerx:


Wardormeur:
Not to be merged before AWS migration is finished

PR #1191, opened 2017-10-03
Remove edit link from users dojos when a lead is completed

Wardormeur:


Wardormeur:
On hold until a decision is taken regarding CDF validation workflow

PR #1199, opened 2017-10-12
Add SSH key fingerprint to use

DanielBrierton:
Turns out we need to explicitly tell Circle which SSH key to use by referencing its fingerprint. Hopefully this is the last modification needed


butlerx:
if we are doing this we could move it back to frontend

PR #1202, opened 2017-10-16
Allow to log params for slow requests

Wardormeur:


Wardormeur:
Just checked NRelics config : requires an update of newrelic package first.

butlerx:
should be possible on staging then?

Wardormeur:
Can't until the package is updated. I'm trying to configure keys for staging atm

PR #1204, opened 2017-10-18
Remove restriction for checkbox on event edition

Wardormeur:
toggle on/off location (if you decide that's a bad idead to check it.. afterwards)
Rename to match backend field
preload of dojo info added for event edition
????

### DIFF
--- a/web/public/js/controllers/dojo-event-form-controller.js
+++ b/web/public/js/controllers/dojo-event-form-controller.js
@@ -117,11 +117,12 @@
 
     $scope.ticketTypesTooltip = $sce.trustAsHtml($scope.ticketTypesTooltip);
 
+    $scope.originalEvent = {};
     $scope.eventInfo = {};
     $scope.eventInfo.dojoId = dojoId;
     $scope.eventInfo.public = true;
     $scope.eventInfo.type = 'one-off';
-    $scope.eventInfo.prefillAddress = true;
+    $scope.eventInfo.useDojoAddress = true;
     $scope.eventInfo.recurringType = 'weekly';
     $scope.eventInfo.sessions = [{name: null, tickets:[{name: null, type: null, quantity: 0}]}];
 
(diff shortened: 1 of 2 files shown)


web/public/js/controllers/dojo-event-form-controller.js
```diff
@@ -260,21 +261,22 @@
...
+      } else if ($scope.originalEvent.city && $scope.originalEvent.address) { // we restore to previous data
+        $scope.eventInfo.city = $scope.originalEvent.city;
+        $scope.eventInfo.address = $scope.originalEvent.address;
       }
-
-      $scope.updateLocalStorage('prefillAddress', false);
-      $scope.eventInfo.city = $scope.eventInfo.address = null;
-      $scope.updateLocalStorage('city', null);
-      $scope.updateLocalStorage('address', null);
+      if (done) return done();
```
butlerx: whats the effect of if done here?

Wardormeur: Usage for both waterfall and front-end.
That's why in the waterfall it's the only function using $scope

web/public/js/directives/event/cd-create-event/template.dust
```diff
@@ -119,16 +119,17 @@
...
+          <input type="checkbox" id="prefill-address" ng-model="eventInfo.useDojoAddress"
+            ng-click="prefillDojoAddress()" ng-disabled="pastEvent"/>
           <label for="prefill-address">{@i18n key="Use Dojo listing address"/}</label>
           <i uib-tooltip="{@i18n key="Use Dojos listing address. This will pre-fill the address fields."/}" tooltip-placement="top" tooltip-trigger="click" class="info-button fa fa-info-circle event-tooltip"></i>
         </div>
       </div>
       <div class="form-group" ng-class="{'has-error': eventForm.$submitted && eventForm.city.$error.required}">
         <label for="city">{@i18n key="City"/}*:</label>
-        <ui-select ng-model="eventInfo.city" ng-disabled="pastEvent" ng-readonly="eventInfo.prefillAddress" name="city" id="city" ng-change="updateLocalStorage('city', eventInfo.city); eventInfo.address = null" required>
+        <ui-select ng-model="eventInfo.city" ng-disabled="pastEvent || eventInfo.useDojoAddress" ng-readonly="eventInfo.useDojoAddress" name="city" id="city" ng-change="updateLocalStorage('city', eventInfo.city); eventInfo.address = null" required>
```
DanielBrierton: The `ng-readonly` here is pointless as the `ng-disabled` will override it as it's a superset of restrictions. I'd be inclined to just get rid of the `ng-readonly` here.

PR #1214, opened 2017-11-10
Track specific event happening in different flows

Wardormeur:


web/public/js/controllers/session-modal-controller.js
```diff
@@ -110,6 +110,7 @@
         'pending': 'Your childs ticket request for %1$s is pending approval'
       };
       //bulkApplyApplications handles the creation of payloads for the email(s) and sends them
+      Analytics.trackEvent($state.current.name, 'click', 'book_tickets');
```
DanielBrierton: Just realised, `$state.current.name` won't match up with `this.$route.name` over in the Vue app, which means we'll end up with the same metric in two different categories rather than just seeing one metric go up/down. Perhaps we should just hard code the Vue route name in here since cp-zen-frontend is the future and this is just a quick monkey patch?

Wardormeur: That's fine for this scenario because we're tracking differently recurring users from new users.
My understanding of the previous GA is that it's nice as a reference over time (everything is caught), but way too "catch-all" for long time, specific, analysis.

PR #1227, opened 2018-01-02
use cp-log-lib rather then console.log

butlerx:
unhandled errors would now be logged in logentries


ArayB:
Closing as this is very old.

PR #1235, opened 2018-02-08
Setup sitemap for dojos urls and some static pages

Wardormeur:


Wardormeur:
integrated into mastermind https://github.com/CoderDojo/cp-zen-platform/pull/1236

PR #1236, opened 2018-02-14
Setup mastermind

Wardormeur:
Setup of the orchestrator/mediator to avoid the cross-µs queries


package.json
```diff
@@ -35,6 +35,7 @@
     "chairo-cache": "0.5.2",
     "country-language": "^0.1.7",
     "cp-i18n-lib": "git+https://github.com/CoderDojo/cp-i18n-lib.git#1.0.0",
+    "cp-permissions-plugin": "https://github.com/CoderDojo/cp-permissions-plugin.git",
```
DanielBrierton: Set version to 1.0.5

web/lib/cp-permissions-pre-handler.js
```diff
@@ -0,0 +1,51 @@
...
+        request.log(
+          ['error', '50x'],
+          {
+            status: '403',
+            host: request.app.hostUid,
+            payload: request.payload,
+            params: request.params,
+            url: request.url,
+            user: request.user,
+            error: response,
```
Wardormeur: shouldn't this be error rather than response?

web/lib/mastermind.js
```diff
@@ -0,0 +1,25 @@
+const { eachSeries } = require('async');
+const { partialRight } = require('lodash');
+const Boom = require('boom');
+
+const stepHandler = (step, cb, req, reply) => step(req, reply, cb);
+
+const errorHandler = (err, reply) => {
+  if (err) return reply(Boom.boomify(err));
```
Wardormeur: nearly tempted if we have access to it (not sure) to add a either a log (to detect "impossible" scenarios) or a reply.continue

web/lib/plugins/sitemap.js
```diff
@@ -0,0 +1,93 @@
...
+  const host = process.env.HOSTNAME || '127.0.0.1';
+  const protocol = process.env.PROTOCOL || 'http';
+  const hostWithPort = `${protocol}://${host}:${port}`;
+  let currentSitemap;
+  const skeleton = {
+    hostname: hostWithPort,
+    cacheTime: 0, // We manage the reset of the cache manually
+    urls: [],
+  };
+  const staticUrls = [
```
Wardormeur: This could be generated at first load of the application through a configuration (but should be kept between each refresh)
To add as a comment ?

web/lib/seneca-web-error-handler.js
```diff
@@ -0,0 +1,14 @@
...
+  const resp = req.response;
+  let statusCode = 200;
+  // This only happens because seneca respond with 200 or 500 only
+  // We shouldn't need it in other microservices which are properly using the HTTP standard
+  if (resp && resp.http$) {
+    if (resp.http$.status) statusCode = resp.http$.status;
+    if (resp.http$.redirect) return reply.redirect(resp.http$.redirect);
+    return reply(Boom.boomify(resp.data, { statusCode }));
+  }
+  return reply(resp);
```
Wardormeur: Some doubts here again (and it's my code again :( )
```
reply - the reply interface which is used to return control back to the framework. To continue normal execution of the request lifecycle, reply.continue() must be called. If the extension type is 'onPostHandler' or 'onPreResponse', a single argument passed to reply.continue() will override the current set response (including all headers) but will not stop the request lifecycle execution. To abort processing and return a response to the client, call reply(value) where value is an error or any other valid response.

```

web/lib/cp-permissions-pre-handler.js
DanielBrierton: That sounds like it makes sense. We've been logging response for quite some time though. 😅

Wardormeur: :x

web/lib/mastermind.js
DanielBrierton: Not quite sure I get what you mean here.

Wardormeur: non-erronous final callback should not get there

web/lib/plugins/sitemap.js
DanielBrierton: How about

web/lib/seneca-web-error-handler.js
DanielBrierton: So, you're saying we should use `reply.continue()` instead?

There's nothing left in our chain of handlers at this point though. This is to pick up any errors that result in no reply and send the error to the client, isn't it?

Wardormeur:
LGTM

PR #1242, opened 2018-02-21
Add Full up error message when there is no space left to accept the i…

Wardormeur:
…nvitation

### DIFF
--- a/web/public/js/controllers/accept-session-invite-controller.js
+++ b/web/public/js/controllers/accept-session-invite-controller.js
@@ -29,8 +29,9 @@
         $state.go('dojo-list');
       }
     }, function (err) {
-      if(err) {
-        console.error(err);
+      if (err.status === 410) {
+        alertService.showError($translate.instant('Full up'));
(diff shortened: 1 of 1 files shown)


web/public/js/controllers/accept-session-invite-controller.js
```diff
@@ -29,8 +29,9 @@
         $state.go('dojo-list');
       }
     }, function (err) {
-      if(err) {
-        console.error(err);
+      if (err.status === 410) {
+        alertService.showError($translate.instant('Full up'));
```
DanielBrierton: Definitely needs a more detailed error.

`This event is fully booked. As a result, you can no longer accept your invitation.`?

/cc @conoramurphy

Wardormeur: Yeha, reused the one used for the event booking to avoid a new translation string

DanielBrierton:
(will need cp-translations update after it's built... We really need to do something about that...)

PR #1243, opened 2018-02-23
Set the env vars prefixed with ZEN_

Wardormeur:


test/lib/plugins/sitemap.spec.js
```diff
@@ -27,7 +27,7 @@ lab.describe('sitemap', () => {
     { urlSlug: 'fr/Lyon' },
     { urlSlug: 'ie/Dublin' },
   ];
-  const host = `${process.env.PROTOCOL || 'http'}://${process.env.HOSTNAME || 'localhost'}:${process.env.PORT || 8000}`;
+  const host = 'http://127.0.0.1:8000';
```
butlerx: hardcoded http and address

butlerx: true

PR #1245, opened 2018-03-08
Enhancement/v3 event service

Wardormeur:
## Description

set a "manual" proxy with validation for event service
set a default reusable transport
set a default reusable query validation

## Motivation and Context

https://trello.com/c/x59K257I

## How Has This Been Tested?

Unit tests + manual tests

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [ ] I gave a thought about security (XSS, perms, injection of params, scoping, ...)
- [x] Linting passes

## Reviewer checklist:

 - [ ] Security
   - [x] Permissions
   - [x] Data scope and leaks
   - [x] Injection
   - [x] XSS
 - [ ] Performance
   - [x] tested with production data
 - [ ] Code
   - [x] Single responsibility
   - [x] Naming
 - [ ] Design
   - [ ] Responsive
   - [ ] Scale well with production data
   - [ ] Looks great !


web/lib/handlers/event.js
```diff
@@ -0,0 +1,17 @@
+const mastermind = require('../mastermind');
+const Event = require('../models/event');
+
+const get = params => // eslint-disable-line no-unused-vars
+  mastermind([
+    // eslint-disable-next-line no-unused-vars
+    async (req, reply, cb) => {
```
butlerx: using async without a try catch is deprecated in node 

https://nodejs.org/dist/latest-v8.x/docs/api/deprecations.html#deprecations_dep0018_unhandled_promise_rejections

Wardormeur: it's caught by mastermind, which is an async series

butlerx: This https://github.com/CoderDojo/cp-zen-platform/blob/master/web/lib/mastermind.js?

I can't see how that would handle a rejected promise?

Wardormeur: See https://caolan.github.io/async/index.html "Using ES2017 async functions"

butlerx: thats was added in async 2.3 though this project still uses 0.9

Wardormeur: Good catch

butlerx: https://github.com/caolan/async/blob/v1.5.2/README.md#asyncify could work if you dont want to upgrade async  all the way to v2

PR #1252, opened 2018-05-11
add orders api to mastermind - currently for creating an order

heyitsmaimai:


web/api/mastermind/order.js
```diff
@@ -0,0 +1,72 @@
...
+const { applications } = validation.definitions;
+
+module.exports = [
+  {
+    method: 'GET',
+    path: `${basePath}/events/{eventId}/orders/{orderId}`,
+    handler: orderHandlers.get(),
+    config: {
+      auth: auth.apiUser,
+      description: 'List A User\'s Order',
```
Wardormeur: *load, not list :)
Alos, it's the current user, not "A" user, as you only load the current user info

web/api/mastermind/order.js
```diff
@@ -0,0 +1,72 @@
...
+      },
+    },
+  },
+  {
+    method: 'POST',
+    path: `${basePath}/events/{eventId}/orders`,
+    handler: orderHandlers.post(),
+    config: {
+      auth: auth.apiUser,
+      description: 'Add A User\'s Order',
```
Wardormeur: Add an order :) It's from an user, but everything is :p

web/api/mastermind/order.js
```diff
@@ -0,0 +1,72 @@
...
+  {
+    method: 'GET',
+    path: `${basePath}/events/{eventId}/orders/{orderId}`,
+    handler: orderHandlers.get(),
+    config: {
+      auth: auth.apiUser,
+      description: 'List A User\'s Order',
+      notes: 'User\'s Order',
+      tags: ['api', 'orders'],
+      plugins: {
```
Wardormeur: Missing permissions? 
We don't want to allow everybody to list everybody's orders (which we don't because it's based on the current user Id), yet, an o13 should be allowed to see he's already booked in by its parent

web/lib/handlers/order.js
```diff
@@ -0,0 +1,33 @@
+const { asyncify } = require('async');
+const mastermind = require('../mastermind');
+const Order = require('../models/order');
+
+const get = params => // eslint-disable-line no-unused-vars
+  mastermind([
+    // eslint-disable-next-line no-unused-vars
+    asyncify(async (req, reply, cb) => {
+      const { userId } = req.user.id;
```
Wardormeur: shouldn't it be `const userId = req.user.id` ? To my knowledge, req.user.id is already a string, using destructuration doesn't help

web/lib/handlers/order.js
```diff
@@ -0,0 +1,33 @@
...
+      const orders = await Order.get(query);
+      return reply(orders).code(200);
+    }),
+  ]);
+
+const post = params => // eslint-disable-line no-unused-vars
+  mastermind([
+    // eslint-disable-next-line no-unused-vars
+    asyncify(async (req, reply, cb) => {
+      const userId = req.user.user.id;
```
Wardormeur: Is it req.user.user.id or req.user.id ? https://github.com/CoderDojo/cp-zen-platform/pull/1252/files#diff-d35579e7e78f0621347d79ee5bfe4725R9

heyitsmaimai: sorry yeah its req.user.user.id. didnt change it for the get

web/api/mastermind/order.js
heyitsmaimai: I see. I have to make a new permission for this?

Wardormeur: Mmmh. We don't currently have a use case for listing orders.
So maybe let it be until then cc @DanielBrierton  ?

DanielBrierton: We have a use case for getting a specific order – the confirmation page.

Wardormeur: Mmh, ok, but it won't be accessible to the children being booked. So it doesn't apply in this scenario, so it could be alright in the current state

DanielBrierton: Well, we'll need to confirm that the user requesting it is the creator. But I think the handler is using the userId of the logged in user, so I guess it should be fine?

Wardormeur: Yup, nothing to see here, skip :)
Thanks !

web/api/mastermind/order.js
```diff
@@ -0,0 +1,72 @@
...
+const validation = require('../validations/orders');
+const orderHandlers = require('../../lib/handlers/order');
+
+const basePath = '/api/3.0';
+const { applications } = validation.definitions;
+
+module.exports = [
+  {
+    method: 'GET',
+    path: `${basePath}/events/{eventId}/orders/{orderId}`,
```
Wardormeur: May I request to remove the {orderId} ? As discussed yesterday, there should be only 1 order per user per event, so that load, if not done in the context of a dojo-champion, should only return the user's order

DanielBrierton: I'm a bit unsure how exactly we should handle this one. `/events/{eventId}/orders` should return all orders for an event that you're allowed to see. You may be a ticketing admin, but still book a ticket for your child, so one endpoint doesn't work.

Part of my wants to say `/events/{eventId}/my-order` to go along with it, but I'm not a big of that. Other options:

Wardormeur: Yet the code behind makes use of the current userId to list those orders.
Either we keep it and it becomes the list of orders that the user can see for this event (which becomes problematic if he's himself booked as well) or 
We use the different endpoint, in which case I'd be happy we remove the "new" one as we have to look into permissions (that I don't want to go into now)
I believe the best one is probably `/events/{eventId}/just-give-me-my-orders-please`, but i'd prefer to avoid "prefixing" the api with `my` (considering the way our routes are being created). 
Another solution would be to reverse the dependency then, like you did for `/users/{userId}/events/{eventId}/orders` but i'd be more inclined to reduce it to `/users/{userId}/orders?eventId=`

DanielBrierton: Actually, yeah, I like the idea of `/users/{userId}/orders?eventId={eventId}` the most because then you can get all orders for a user by just omitting the eventId, which could be useful for a future version of my tickets.

DanielBrierton: I also like the idea of allowing the URL to be `/users/me/orders` so you don't have to bother getting the logged in user's ID ;)

Wardormeur: What if `me` is a valid GUID ? :taco:

Wardormeur: @heyitsmaimai Url of `events/{eventId}/order/{orderId}` to be changed to `users/{userId}/orders` with optional query param `eventId`

DanielBrierton: Actually, I assume it should really be `?query[eventId]=asdf`?

Wardormeur:
Missing tests

PR #1254, opened 2018-05-18
Replace next with typoed version which is supported by both ng and Vue

Wardormeur:


DanielBrierton:
@Wardormeur Superceded and can be closed, right?

Wardormeur:
Nope.

PR #1255, opened 2018-05-30
Send email post-ordering

Wardormeur:
WIP, waiting for email templates to assign proper values

### DIFF
--- a/web/lib/models/email.js
+++ b/web/lib/models/email.js
@@ -1,0 +1,22 @@
+const service = process.env.EMAIL_SERVICE;
+const Transport = require('../transports/http');
+const moment = require('moment');
+
+class Email {
+  constructor() {
+    this.transport = new Transport({
+      baseUrl: `http://${service}:3000/`,
+      json: true,
+    });
+  }
+  formatTime(date) {
+    return moment.utc(date).format('HH:mm');
+  }
+  formatDate(date) {
+    return moment.utc(date).format('Do MMMM YY');
+  }
+  post(body) {
+    return this.transport.post('email/send', { body });
+  }
+}
+module.exports = Email; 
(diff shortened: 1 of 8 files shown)


web/lib/models/event-emails.js
```diff
@@ -0,0 +1,75 @@
+const Email = require('./email');
+
+const port = process.env.ZEN_PORT || 8000;
+const host = process.env.ZEN_HOSTNAME || 'localhost';
+const protocol = process.env.ZEN_PROTOCOL || 'http';
+
+class EventEmail extends Email {
+  formatEventDate(event) {
+    if (event.type === 'recurring') {
+      return `${this.formatDate(event.dates[0].startTime)} - ${this.formatDate(event.dates[event.dates.length - 1].startTime)}`;
```
DanielBrierton: Not a fan of this... Suggests it's on every day instead of every week/two weeks/month. Should probably be the same as frontend..?

web/lib/models/email.js
```diff
@@ -0,0 +1,24 @@
...
+      json: true,
+    });
+  }
+  // eslint-disable-next-line class-methods-use-this
+  formatTime(date) {
+    return moment.utc(date).format('HH:mm');
+  }
+  // eslint-disable-next-line class-methods-use-this
+  formatDate(date) {
+    return moment.utc(date).format('Do MMMM YY');
```
DanielBrierton: i18n

web/lib/models/event-emails.js
Wardormeur: nuxt?

web/lib/models/email.js
Wardormeur: No.

PR #1257, opened 2018-06-03
Feature/checkin order

Wardormeur:
Missing : 
 - test, 
 - refine endpoints


Wardormeur:
https://github.com/CoderDojo/cp-zen-frontend/pull/133
https://github.com/CoderDojo/events-service/pull/14

Wardormeur:
TODO as well : 
- contact http://goqr.me/api/doc/create-qr-code/#general_tos to not be considered as spam for demo purpose.
Move to s3/cached in a way once we know it's used.

DanielBrierton:
Why use a QR code service instead of just using a library for it? There's a few libraries that can do frontend generation, and a few that do backend (QRious does both, but in two separate packages...). The only time we'd need the backend option is for emails, otherwise just frontend render everything so we don't have to worry about image generation too much on the server.

Wardormeur:
It's a demo. Uptimately, i'd prefer the qrcode to be static (image stored)

Wardormeur:
Also, a lib for something that is used in one page is a bad reason to increase your .js size files, that I'd like to avoid :)

DanielBrierton:
Code split to avoid it loading on all other pages then. But yeah, we could just statically store them I guess. Generate, upload to S3 and forget about it.

Wardormeur:
TBD : add unit tests

Wardormeur:
TBD: 
 - review permissions/security (response by API contains PII through the applications)
- change API endpoint from /checkin to /attendances
- eventually avoid generated files through a published version of qr-code-service

PR #1262, opened 2018-06-26
Send an email after updating an order

Wardormeur:
Add replyTo and sendgrid category tags

### DIFF
--- a/web/lib/handlers/order.js
+++ b/web/lib/handlers/order.js
@@ -27,7 +27,8 @@ const post = params => // eslint-disable-line no-unused-vars
       reply(req.app.order).code(200);
       return next();
     },
-  ].concat(sendBookingEmail));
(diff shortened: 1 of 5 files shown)


web/lib/handlers/order.js
```diff
@@ -27,6 +27,21 @@ const post = params => // eslint-disable-line no-unused-vars
       reply(req.app.order).code(200);
       return next();
     },
+  ].concat(sendBookingEmail));
```
DanielBrierton: Super minor, but I'd rather destructuring here.
```
      reply(req.app.order).code(200);
      return next();
    },
    ...sendBookingEmail,
  ]
```
I think it just looks cleaner and a bit more obvious.

Wardormeur: oky :]

PR #1277, opened 2018-08-16
Change SAD's firstSession into a dropdown

Wardormeur:
Remove firstSession formatting as date
Make it optional so it doesn't screw up with previous leads having a different format


Wardormeur:
Requires cp-translations

Wardormeur:
https://github.com/CoderDojo/cp-translations/pull/83

PR #1296, opened 2018-10-09
Reduce possible params to what found in codebase

Wardormeur:


web/api/dojos.js
```diff
@@ -241,21 +240,30 @@ exports.register = function (server, eOptions, next) {
       },
       cors: { origin: ['*'], credentials: false },
       validate: {
-        payload: Joi.alternatives().try(Joi.any(), Joi.object({ query: {
+        payload: Joi.object({ query: {
```
DanielBrierton: Check if people integrating with us use anything other than fields listed.

Wardormeur: Seems alright : https://github.com/CoderDojoNederland/Website/blob/e9943ecc7b990a3ad6615e44e36a03aad4ed4409/src/CoderDojo/CliBundle/Service/ZenApiService.php#L38

PR #1297, opened 2018-10-10
Limit api uses-cases to know ones

Wardormeur:


web/api/users.js
```diff
@@ -318,6 +348,13 @@ exports.register = function (server, eOptions, next) {
           ],
         },
       },
+      validate: {
+        payload: {
+          token: Joi.string().guid(),
```
DanielBrierton: These should all be `.required()`, right?

Wardormeur: yay should :+1:

PR #1299, opened 2018-10-11
Set api validation for multiples endpoints

Wardormeur:


web/api/badges.js
```diff
@@ -79,6 +89,12 @@ exports.register = function (server, eOptions, next) {
       auth: auth.apiUser,
       description: 'Request a badge',
       tags: ['api', 'users'],
+      validate: {
```
DanielBrierton: Isn't this actually an endpoint that can just be removed? Badges are auto-awarded for some time now. No need to claim them.

web/api/profiles.js
```diff
@@ -196,6 +196,14 @@ exports.register = function (server, eOptions, next) {
       auth: auth.apiUser,
       description: 'Invite child',
       tags: ['api', 'users'],
+      validate: {
+        payload: {
+          ninjaData: {
+            ninjaEmail: Joi.string().email(),
```
DanielBrierton: required?

Wardormeur: yes, will do :+1:

PR #1310, opened 2018-10-16
Add support for register for FTB

Wardormeur:


Wardormeur:
Was on outdated cp-zen-frontend, fixed

PR #1315, opened 2018-10-18
Remove promote once and for all

Wardormeur:


web/api/users.js
```diff
@@ -260,22 +260,6 @@ exports.register = function (server, eOptions, next) {
         }),
       },
     },
-  }, {
```
DanielBrierton: 👍

butlerx: Awh my fav privilege escalation

PR #1317, opened 2018-10-22
Ensure every endpoint has validation

Wardormeur:
Set validation abomination with controller rework for update user/youth

### DIFF
--- a/web/api/badges.js
+++ b/web/api/badges.js
@@ -61,7 +61,7 @@ exports.register = function (server, eOptions, next) {
       tags: ['api', 'users'],
       validate: {
         params: {
-          userId: Joi.string().required(),
+          userId: Joi.string().guid().required(),
         },
       },
     },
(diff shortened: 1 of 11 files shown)


Wardormeur:
List is not finished alas

web/api/badges.js
```diff
@@ -54,6 +59,11 @@ exports.register = function (server, eOptions, next) {
       auth: auth.apiUser,
       description: 'Load user badges',
       tags: ['api', 'users'],
+      validate: {
+        params: {
+          userId: Joi.string().required(),
```
DanielBrierton: `.guid()`

web/api/dojos.js
```diff
@@ -888,6 +891,8 @@ exports.register = function (server, eOptions, next) {
       validate: {
         params: {
           dojoId: Joi.string().required(),
```
DanielBrierton: `.guid()`

PR #1327, opened 2018-11-08
Restrict API payloads for events

Wardormeur:
Remove some unknowns
Fix the 'empty' calls that was breaking the API doc


Wardormeur:
Requires https://github.com/CoderDojo/cp-events-service/pull/191

PR #1333, opened 2019-02-05
Expose an endpoint for user deletion

Wardormeur:
Connect to multiple new backends (users & clubs)
Expose the cdf display for deletion of users through angular redirection to vueJs

### DIFF
--- a/web/api/mastermind/user.js
+++ b/web/api/mastermind/user.js
@@ -1,0 +1,88 @@
+const Joi = require('joi');
+const auth = require('../../lib/authentications');
+const validation = require('../validations/users');
+const userHandlers = require('../../lib/handlers/user');
+
+const basePath = '/api/3.0';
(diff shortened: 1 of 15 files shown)


web/api/mastermind/user.js
```diff
@@ -0,0 +1,88 @@
...
+          responseMessages: [
+            { code: 400, message: 'Bad Request' },
+            { code: 204, message: 'OK' },
+          ],
+        },
+      },
+      validate: {
+        params: {
+          userId: Joi.string().guid().required(),
+          soft: Joi.boolean(),
```
DanielBrierton: Shouldn't `soft` and `cascade` be under `query` instead of `params`?

Wardormeur: payload even. Oof.
Good catch

PR #1337, opened 2019-02-12
Send state and county to the backend during dojo/lead update

Wardormeur:


Wardormeur:
TBD: 
- [ ] add support for single administrative_code
- [ ] add support for region short_code
- [ ] add support for city

PR #1339, opened 2019-02-14
Feature/1236 amend referer answers

ArayB:
As per [Community platform Issue 1236](https://github.com/CoderDojo/community-platform/issues/1236) amend the options in the "How did you hear about CoderDojo?" select when creating a Dojo.


Wardormeur:
Looks like I've never finished to migrate this... The reference is also used by the cdf panel (/dashboard/manage-dojos -> click on "Preview" for a Dojo/Lead) https://github.com/CoderDojo/cp-zen-platform/blob/master/web/public/js/services/dojo-utils.js#L134

ArayB:
The linting has really done a number on this.
Will close and do a separate pr without all the formatting changes

PR #1340, opened 2019-02-15
Feature/1236 referer amends

ArayB:
Add new options to the "How did you hear about CoderDojo?" select when creating a new Dojo as per [community platform issue 1236](https://github.com/CoderDojo/community-platform/issues/1236)

Ensure removed options are still visible on admin pages.

Add prettier to eslint to make all the things pretty.


web/public/js/services/dojo-utils.js
```diff
@@ -135,24 +135,56 @@ angular.module('cpZenPlatform').factory('dojoUtils', ['$location', '$translate',
...
+      name: $translate.instant('Event / Conference / Training')},
+    { id: 'library',
+      name: $translate.instant('Library or youth group')},
     { id: 'word_of_mouth',
       name: $translate.instant('Word of Mouth')},
+    { id: 'other',
+      name: $translate.instant('Other (comment field)')}
+  ];
+
+  dojoUtils.startingDojoSrcsWithDeprecated = [
```
Wardormeur: Wouldn't it be easier to extend this list with startingDojoSrcs rather than duplicating the definition ?

web/public/js/services/dojo-utils.js
```diff
@@ -135,24 +135,56 @@ angular.module('cpZenPlatform').factory('dojoUtils', ['$location', '$translate',
...
+    { id: 'foundation',
+      name: $translate.instant('Raspberry Pi Foundation / CoderDojo Foundation outreach')},
+    { id: 'social',
+      name: $translate.instant('Social media')},
     { id: 'future_learn',
       name: 'Future Learn'},
-    { id: 'developpers',
-      name: $translate.instant('Development Community')},
+    { id: 'employer',
+      name: $translate.instant('My employer (CSR scheme at work)')},
```
Wardormeur: For any new string, or modified string that is translated, you need to modify https://github.com/CoderDojo/cp-translations/blob/master/strings/en_US/messages.po

web/public/js/services/dojo-utils.js
ArayB: 🤦‍♂️

web/public/js/services/dojo-utils.js
ArayB: Ah ok 👍

ArayB: Do I need to update these as well?
https://github.com/CoderDojo/cp-zen-platform/blob/master/web/public/js/services/translation-keys.js#L158

Wardormeur: The idea behind the translationKeys was to not tie the text of the key to the frontend, so we could change the translation when necessary without touching the frontend.
We didn't use them much in the end (and changed frontend framework...)
So nope, don't worry about it

PR #1344, opened 2019-02-26
Upgrade dustjs due to ReDos

Wardormeur:


Wardormeur:
We're trading a high severity for 2 low, that's still worth it

PR #1359, opened 2019-04-08
Fix vulnerable version of js-yaml

Wardormeur:
Thanks god for the 'resolutions' block that allows us to bypass intermediate upgrades, at the risk of instability


grega:
What @dtt101 said!

PR #1362, opened 2019-04-23
Add filtering by name to manage-users

Wardormeur:
as a param, you can now link to a specific search
Add a message to change tab if there is a conflict on accepting a join request


Wardormeur:
Requires https://github.com/CoderDojo/cp-translations/pull/106

PR #1363, opened 2019-05-08
Redirect event form to the Vuejs one by default

Wardormeur:
Partially solves https://github.com/CoderDojo/community-platform/issues/1260


ArayB:
This has been covered by #1372

PR #1366, opened 2019-05-14
Remove international phone validation library

ArayB:
The library was preventing valid numbers for Brazilian mobile phones
being added.
The library has not been updated in 2 years so unlikely to be fixed.

Removing phone number validation as trying to validate phone number
format is incredibly difficult with all the different international
formats possible.

### DIFF
--- a/web/public/js/directives/cd-start-dojo/champion/template.dust
+++ b/web/public/js/directives/cd-start-dojo/champion/template.dust
@@ -50,11 +50,9 @@
     <h3>{@i18n key="Contact details to help us verify your Dojo"/}</h3>
     <div class="form-group">
       <label for="phone">{@i18n key="Phone number"/}</label>
-      <input type="tel" name="phone" id="phone" autocomplete="off" class="form-control"
-        utils-script="/components/intl-tel-input/lib/libphonenumber/build/utils.js"
-         required cd-phone-validator international-phone-number ng-model="$ctrl.champion.phone">
+      <input type="tel" name="phone" id="phone" class="form-control" required ng-model="$ctrl.champion.phone">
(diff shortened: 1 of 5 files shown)


web/public/js/directives/cd-start-dojo/champion/template.dust
```diff
@@ -50,11 +50,9 @@
     <h3>{@i18n key="Contact details to help us verify your Dojo"/}</h3>
     <div class="form-group">
       <label for="phone">{@i18n key="Phone number"/}</label>
-      <input type="tel" name="phone" id="phone" autocomplete="off" class="form-control"
-        utils-script="/components/intl-tel-input/lib/libphonenumber/build/utils.js"
-         required cd-phone-validator international-phone-number ng-model="$ctrl.champion.phone">
+      <input type="tel" name="phone" id="phone" class="form-control" required ng-model="$ctrl.champion.phone">
         <div ng-messages="$ctrl.championForm.phone.$error" ng-show="$ctrl.championForm.phone.$touched && !$ctrl.championForm.phone.$valid">
-          <label class="control-label has-error validationMessage">{@i18n key="Telephone number is not valid"/}</label>
```
Wardormeur: That means cp-translation should be updated. The key for that translation changed :)

Wardormeur:
Pretty sure that injection should be removed as well https://github.com/CoderDojo/cp-zen-platform/blob/master/web/public/js/cp-zen-platform.js#L28

web/public/js/directives/cd-start-dojo/champion/template.dust
ArayB: The "Telephone Number is empty" key already exists in cp-translations, assuming the `strings/en_US/messages.po` is the right place to check

Wardormeur: Apologies, all good then !

PR #1375, opened 2019-10-20
removing exports as they no longer work with latest docker-compose

pbackx:
This is a small fix for an issue that recently popped up due to a change in docker-compose:
https://github.com/docker/compose/issues/6511

Basically, "export var=val" is no longer allowed in  .env files. It used to work, but mostly by accident. The proper syntax is "var=val"

My initial report of this issue can be found here: https://github.com/CoderDojo/cp-local-development/issues/83


ArayB:
Hey, closing this as it seems to have been covered by later work.

PR #1386, opened 2020-01-20
Oauth flow error handling

Jeddf:
- Oauth user error display - use new error display query param in zen frontend
- Error logging on local dev (currently too noisy to turn on in prod)
- Temporarily verbose client side messages while oauth flow is private

Relates to item in stage 2 of https://github.com/CoderDojo/community-platform/issues/1290

### DIFF
--- a/web/config/options.js
+++ b/web/config/options.js
@@ -27,6 +27,10 @@ module.exports = {
   },
 
   hapi: {
+    debug:
+      process.env.NODE_ENV === 'development'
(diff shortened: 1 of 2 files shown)


Jeddf:
> LGTM.
> 
> As these errors are user facing we will need to get the error messages into cp-translations at some point.

Agreed!

I had the thought that having someone be able to have our site show any error message to the user by redirecting them to our site with a query param is bad too so will restrict that ability before final release also.

PR #1388, opened 2020-01-22
Whitelist referer query param

grega:
Fixes: https://github.com/CoderDojo/community-platform/issues/1292


web/controllers/home/index.js
```diff
@@ -1,254 +1,255 @@
...
+      reply.view("index", request.app);
     },
     config: {
       validate: {
         query: {
           profileAuth: Joi.bool(),
-        },
-      },
-    },
+          referer: Joi.string()
```
grega: This is the actual addition

Issue #1387, opened 2020-01-21
RPI profile edit & update


Jeddf:
Stage 2 task from Profile RPI migration epic CoderDojo/community-platform#1290

### What
Create path for zen users to edit their linked my.raspberrypi profile.

### Why
my.raspberrypi should be the only editing page for zen profile information that is not specific to zen.

### How
- Add /rpi/edit endpoint that redirects to rpi profile edit page
- Add returnTo query param capablity to rpi profile edit page
- Have /rpi/login endpoint update stored zen user profile from rpi user profile if profileId is known
- If user session has profileId, serve zen edit profile page with a link to rpi edit profile and just the zen specific info available to edit.

PR #1389, opened 2020-01-22
Update Zen Account on RPI Login

Jeddf:
- `/rpi/edit` route for editing rpi profile
- Zen profile update on rpi login with known account

Partially resolves issue https://github.com/CoderDojo/cp-zen-platform/issues/1387
returnTo on profile edit is dependent on https://github.com/RaspberryPiFoundation/profile/pull/1027

### DIFF
--- a/web/lib/rpi-auth.js
+++ b/web/lib/rpi-auth.js
@@ -2,24 +2,25 @@ const oauth2 = require('simple-oauth2');
 const jwt = require('jsonwebtoken');
 const { URLSearchParams } = require('url');
 
-const homeServer = process.env.HOSTED_URL;
-const loginPath = '/rpi/login';
-const callbackPath = '/rpi/cb';
 const authServer = process.env.RPI_AUTH_URL;
-const authPath = '/oauth2/auth';
+const homeServer = process.env.HOSTED_URL;
 const tokenServer = process.env.RPI_TOKEN_URL || authServer;
-const tokenPath = '/oauth2/token';
 const profileServer = process.env.RPI_PROFILE_URL;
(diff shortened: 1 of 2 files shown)


web/lib/rpi-auth.js
```diff
@@ -42,11 +43,18 @@ function getLogoutRedirectUri() {
 
 function getRegisterRedirectUri() {
   const params = new URLSearchParams();
-  params.set('brand', 'coderdojo');
+  params.set('brand', brand);
```
grega: I might well be missing something, but where is `brand` being set?

web/lib/rpi-auth.js
```diff
@@ -42,11 +43,18 @@ function getLogoutRedirectUri() {
...
   params.set('returnTo', `${homeServer}${loginPath}`);
   return `${profileServer}${profileSignupPath}?${params}`;
 }
 
+function getEditRedirectUri() {
+  const params = new URLSearchParams();
+  params.set('brand', brand);
```
grega: Same as above.

web/lib/rpi-auth.js
Jeddf: It's just a string const being set up on line 18

web/lib/rpi-auth.js
Jeddf: Again up on L18, I'm gonna clean up that block of variables up there, I've left 'em very messy!

web/lib/rpi-auth.js
grega: Ha, of course - I should really look at the whole file rather than the diff :)

PR #1408, opened 2020-05-22
Add maintenance page

ArayB:
Add a maintenance page that we can use during the account migration process.

We can use cloudflare page rules to send people away from the login and register pages while the migration runs so they view this page instead.

Layout is copied from the 404 page.

Screenshot 2020-05-22 at 13 53 22


patch0:
LGTM :+1:

PR #1418, opened 2020-07-16
Need to add new languages to the config.

glenpike:
Also update yarn.lock for correct packages of cp-translations and cp-frontend.


glenpike:
Caveat:

Currently the text is not right-to-left for this language - we might need to get input from people who can read Arabic to guide us because there is often a mixture of Arabic and English text on the page, so setting this on the html tag: `<html lang='ar' direction="rtl">` is too extreme.  It can be done on an element by element basis, and `auto` can also be used, but this sets direction for the entire container it seems, so `<html lang='ar' direction="auto">` for Arabic will use RTL where it is the most prominent on the page.  https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir & https://developer.mozilla.org/en-US/docs/Web/HTML/Element/bdi

Default:

Screenshot 2020-07-16 at 22 07 31

rtl
Screenshot 2020-07-16 at 22 17 09

glenpike:
I think I've got the language into Zen Frontend now, but need to sort the bits that use Zen (Angular).  This should make sure a fair chunk of the live site reads in Arabic.

DanielBrierton (approved):

PR #1427, opened 2020-11-13
Updated the tranlations and frontend deps for new banner

patch0:
This ensures we're pulling in the latest banner.  I hope!


glenpike (approved):
LGTM

PR #1428, opened 2020-11-13
Update cp-zen-frontend

patch0:
I update the version of cp-zen-frontend in packages.json and re-ran yarn install.


glenpike (approved):
LGTM

PR #1462, opened 2022-04-27
Update translations to pull in new email content

patch0:
This is to pull in the new email content from CoderDojo/cp-translations#146


patch0:
Not actually needed in cp-zen-platform -- the content is in cp-dojos-service.

PR #1470, opened 2022-09-27
Update zen-frontend version

IzzySmillie:
- Updates `cp-zen-frontend` to match new [version](https://github.com/CoderDojo/cp-zen-frontend/blob/master/package.json#L3) 1.0.153


ChristinaVoss (approved):
Probably all good! 😅

PR #1484, opened 2023-04-14
Add join as parent API endpoint

patch0:
This allows a request to be made to update a users roles on a Dojo to include parent-guardian, returning an error if that role already exists:
```
curl --request POST 'http://localhost:8000/api/3.0/dojos/DOJO-ID/join-as-parent-guardian' \
--header 'Authorization: Bearer TOKEN'
{"id":"7a3d718b-18af-4f47-b4f8-c03b7285a5eb","userId":"USER-ID","dojoId":"DOJO-ID","userTypes":["parent-guardian"],"userPermissions":[],"mysqlUserId":null,"mysqlDojoId":null,"owner":null,"backgroundChecked":false,"deleted":0,"deletedBy":null,"deletedAt":null}
```

And then, if the same POST is made again:
```
curl --request POST 'http://localhost:8000/api/3.0/dojos/DOJO-ID/join-as-parent-guardian' \
--header 'Authorization: Bearer TOKEN'
{"statusCode":400,"error":"Bad Request","message":"400 - \"Membership already has this role\""}
```

### DIFF
--- a/web/api/mastermind/dojo.js
+++ b/web/api/mastermind/dojo.js
@@ -188,12 +188,12 @@ module.exports = [
   },
   {
     method: 'POST',
-    path: `${basePath3}/dojos/{id}/join`,
-    handler: membershipHandlers.join(),
+    path: `${basePath3}/dojos/{id}/join-as-parent-guardian`,
+    handler: membershipHandlers.joinAsParentGuardian(),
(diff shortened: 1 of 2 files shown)


peconia (approved):
🚢

PR #1487, opened 2023-05-30
Update build process to use Debian Buster to fix SSL certificate issues

patch0:
When installing packages with bower, recent SSL certificate changes have caused errors like

```
bower CERT_HAS_EXPIRED      Request to https://registry.bower.io/packages/angular-google-maps failed: certificate has expired
error Command failed with exit code 1.
```

which cause both the circleci test process and docker build process to fail.

The solution is to set `NODE_OPTIONS=--use-openssl-ca` in the environment, however this doesn't appear to work for the Alpine builds of node.  It does however work for Debian Buster version.

This PR updates the Dockerfile to start from `node:8-buster-slim` and build from there.  There resulting docker image is just over 1.1GB, but that's the best I could do.  I tried using layers but that didn't make the resulting size smaller.

I have run the built container locally and it all works. 🤞 it works in prod too.. (we can roll back if not)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ jobs:
   build:
     working_directory: ~/cp-zen-platform
     docker:
-      - image: node:carbon
+      - image: node:8-buster
         environment:
           NODE_ENV: testing
           EVENTS_SERVICE: events2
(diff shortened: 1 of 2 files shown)


peconia (approved):
🤞

PR #1489, opened 2023-06-06
Refactor login/logout to share session clearing code

patch0:
I noticed that when logging in with another user on top of an existing session that the old session was still present on the "account select" screen.  This change makes sure the session is cleared before going through the login flow, the same way as it is cleared in the logout flow.

Also I've added an option to the gulpfile to allow large HTTP headers to be parsed, as I was seeing HPE_HEADER_OVERFLOW errors in the HAPI debug logs.  I don't think this will happen in prod as not as many cookies are sent, as more than one domain is used for profile, zen, cd frontend etc.


peconia (approved):
🍅

